PR #3, opened 2026-07-09
Phase 2: daemon-first infrastructure

wesm:
Turns docbank from a direct-vault-access CLI into a daemon-first system, aligning it with msgvault and the other kit-based projects.

## What's here

- **Daemon** (`docbank serve`, `serve start/stop/status [--json]`): one process owns the SQLite store and blob directory under an exclusive flock, acquired non-blocking. Runtime record + `/api/ping` discovery via kit, with a gopsutil create-time guard against PID reuse and a token-gated hidden shutdown route. Background daemons idle out after `idle_timeout`; foreground never does.
- **HTTP API** (huma v2 on the stdlib mux, `/api/v1`): stat/list/content/search reads, dir create, id-addressed move/trash/restore guarded by per-node `If-Match` revisions (428 missing / 400 malformed / 412 stale), transactional path-addressed move/trash, server-side-path ingest (loopback-fenced), trash/gc/verify behind an in-daemon maintenance gate. RFC 7807 error envelope with a `code` member as the single wire contract.
- **CLI cutover**: every data command is an HTTP client with daemon auto-start (version-mismatched daemons are stopped and replaced under a launch lock). `openVault`, the shared/exclusive lock API, and the CLI's direct store access are deleted.
- **Auth**: keyless is loopback-only; non-loopback binds require an API key; public binds are refused outright.
- **`docbank update`**: kit selfupdate with stop→install→restart daemon coordination and rollback-to-old-daemon on install failure; refuses unverified binaries.
- **`docbank openapi`**: emits the OpenAPI document offline (no daemon or store needed).
- **Release workflow**: tag-driven, native CGO runners (linux amd64/arm64, darwin arm64 — macos-13 retirement dropped darwin/amd64), producing the exact asset + SHA256SUMS layout kit selfupdate consumes.
- **Config**: `config.toml` in `$DOCBANK_HOME` (bind, port, key, idle timeout, web toggle) with strict unknown-key rejection.
- **Docs**: zensical site updated — new daemon page, rewritten configuration/locking/HTTP-API/integrity pages, CLI reference, roadmap split into Phase 2a (implemented) / 2b (designed).

## Verification

- `go test -tags fts5 ./...` green, including a real-binary daemon lifecycle integration test; `make lint` and strict `make docs-build` clean; `actionlint`/`zizmor` clean on the release workflow.
- End-to-end smoke: `docbank add` auto-starts the daemon, `ls`/`status`/`/docs` serve, `serve stop` leaves no runtime record.
- Every task passed an independent spec + quality review; the branch passed a whole-branch final review plus roborev checkpoints (all findings fixed or refuted with evidence).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


wesm:
Re the combined review on be3a337:

**Unauthenticated loopback API** — fixed in 7e1e33c + 2de325f. The daemon now always enforces an API key: a configured `api_key` is used as-is; otherwise a per-run key is generated and published to same-user clients through the runtime record inside the 0700 `$DOCBANK_HOME` (the same channel as the existing shutdown token), so other local OS users can read neither the key nor the record. Clients pick the key up during discovery with no UX change; `Ensure` additionally rejects and replaces daemons whose records predate mandatory keys. Verified end-to-end: unkeyed loopback `curl` gets 401, keyed CLI calls succeed, and the key never appears in logs.

**`new("papers")` does not compile** — false positive. `new(expr)` is valid Go as of 1.26 (this repo pins go 1.26.3; toolchain go1.26.4): `go test -tags fts5 ./internal/client/` compiles and passes locally and in CI on all three platforms. No change made.

wesm:
Re the combined review on 2059997 (addressed in 29cd3b2):

**High — plain-HTTP key exposure on non-loopback binds**: real; fixed by taking the first suggested option. `Validate` now rejects every non-loopback bind — keyed or not — since a key in cleartext is not protection, and the docs document SSH tunnel/VPN as the remote path until the daemon grows TLS. The keyed-LAN-bind feature is gone rather than papered over.

**High — `new("papers")` invalid Go**: false positive for the third consecutive review. Value-form `new(v)` is Go 1.26 syntax; go.mod pins 1.26.3 and three-platform CI compiles and passes this exact file. We attempted the suggested `ptr` helper anyway — golangci-lint's modernize fixer auto-inlined it straight back to `new(v)`, which settles the canonical form for this repo. `.roborev.toml` now carries explicit guidance (Go 1.26 features; compilation is CI's job) so this stops recurring.

**Medium — release builds to `dist/` without mkdir**: refuted empirically. `go build -o dist/docbank` on go1.26.4 (the version CI resolves via go-version-file) creates the output directory; verified with a scratch module and a nonexistent `dist/`.

**Medium — `ParseAge` day-count overflow**: real; a count near 213504d wraps past 2⁶⁴ to a small positive cutoff, which would make `trash empty --older-than` delete far newer entries than asked. Day counts that overflow `time.Duration` are now rejected, with regression tests covering the wrap-to-positive case.

wesm:
Response to the combined review of `0875eea`:

**High — unsigned release checksums** (`internal/update/update.go`, `release.yml`): deferred deliberately, on record. The proposed CI-held signing key is inconsistent with this project's threat model: the only party who could swap release assets can compromise the repository itself, and would then also control the signing workflow — signatures add nothing until there is an offline/protected signing key. Wanted, but later: tracked as kata issue `0zdm` ("Sign releases for docbank update") to land before the repo goes public, with implementation notes (kit/selfupdate already verifies ed25519 via `TrustedPublicKeys`). The decision is now documented at the `AllowUnsignedChecksums` site and in `.roborev.toml` guideline 9 (`f220f58`).

**Medium — ingest accepts empty/relative `dest`** (`internal/api/routes_ops.go`): confirmed and fixed in `acf86ff`. An explicit `""` now gets the same `/inbox` default as an absent field (the schema default only covers absence), and a non-absolute dest returns the same 422 `validation` error the path-mutation endpoints use. Covered by new cases in `TestIngestEndpoint`, including a check that an empty-dest import lands at `/inbox/...`, not `/`.

wesm:
Response to the combined review of `f220f58`:

**High — ingest reachable through SSH tunnels** (`middleware.go`, `routes_ops.go`): refuted — there is no principal with the described capabilities. Establishing an SSH tunnel to the daemon host's loopback listener requires an SSH account on that host, which already grants filesystem read as that user — strictly more than ingest exposes. The documented tunnel/VPN guidance is for the vault owner reaching their own daemon; the loopback fence on ingest is a request-shape guard (server-side paths make no sense for remote clients, who get a 403 pointing at planned multipart upload), not a security boundary against key-holders. Recorded in `.roborev.toml` guideline 7 (`cdfaf62`).

**High — unsigned release checksums** (`update.go:35`): adjudicated in the previous round — see the response on `0875eea`, kata issue `0zdm`, and `.roborev.toml` guideline 9 (added by the very commit under review). A CI-held signing key is controlled by the same attacker who could swap assets; signing lands with an offline/protected key before the repo goes public.

**Medium — OpenAPI missing security schemes** (`server.go`): confirmed and fixed in `cdfaf62`. The document now declares `X-Api-Key` (apiKey/header) and `bearer` (http) schemes with a document-level security requirement, so generated clients know credentials are mandatory. Covered by `TestOpenAPIDeclaresSecurity`.

**Medium — misleading trash response paths** (`routes_mutate.go`): confirmed and fixed in `cdfaf62`. Both trash endpoints now capture the pre-trash path before the mutation (safe: the maintenance gate serializes mutations) and their OpenAPI descriptions document that `path` is the pre-trash location a restore would return to. Pinned by new assertions in `TestTrashAndRestoreRoundTripHTTP` and `TestMovePathAndTrashPath`.

wesm:
Response to the review of `cdfaf62`:

**Medium — stale pre-trash path under concurrent mutation**: confirmed and fixed in `7a0cfe0`. The premise of my previous fix was wrong — `g.mutate` takes the RWMutex's shared side (only maintenance ops are exclusive), so ordinary mutations do run concurrently and an ancestor move between the path capture and the trash could stale the response without touching the target's revision. Per the suggested fix, `Store.Trash` and `Store.TrashPath` now compute and return the pre-trash path inside the same transaction that performs the trash; the handlers' separate pre-reads are gone. The path CTE was factored into a tx-capable `pathOf` helper, and the returned path is asserted at the store level in `TestTrashAndRestoreRoundTrip` and `TestTrashPath`.

wesm:
Response to the review of `7a0cfe0`:

**Medium — release build needs `mkdir -p dist`**: refuted, second occurrence (first raised and refuted in the round on the pre-daemon-reshape commits). On the pinned toolchain, `go build -o` creates missing output parent directories — re-verified empirically just now with go1.26.4 against a doubly-nested nonexistent path (`-o new-dir/nested/docbank` succeeds and creates both levels). The release workflow pins the same toolchain via `go-version-file: go.mod`. Now recorded in `.roborev.toml` guideline 2 (`38d4451`) so it stops recurring.

wesm:
Merging this so the work can move forward, it will have to be refined in subsequent PRs

PR #4, opened 2026-07-09
Prepare docbank for v0.1.0

wesm:
Docbank should enter its first tagged release on a stable shared-infrastructure dependency, with destructive intent and bounded-result completeness made explicit before users or external applications depend on those contracts. This branch removes the remaining high-priority release sharp edges and validates the artifact path without creating the tag.

## What changed

- upgrade `go.kenn.io/kit` from the Phase 0 pseudo-version to stable v0.4.0 after auditing the daemon and self-update APIs
- make `trash empty` dry-run by default across store, HTTP API, client, CLI, tests, and docs; require `--run` for permanent deletion
- add `search --limit` (1-1000) and propagate explicit `limit`/`truncated` metadata through the API
- prepare the v0.1.0 changelog and install docs for the existing Linux amd64/arm64 and macOS arm64 archives
- harden CI/release concurrency and release shell interpolation; document the release-write permission

## Verification

- `CGO_ENABLED=1 make test`
- `CGO_ENABLED=1 make lint`
- `make docs-build`
- `go mod verify`
- `prek run`
- `actionlint .github/workflows/*.yml`
- `zizmor --pedantic .` (zero findings)
- local v0.1.0 darwin/arm64 workflow smoke: version metadata, archive naming/content, and SHA256SUMS verification

After this is merged, pushing `v0.1.0` from `main` is the remaining release action; this PR does not merge or tag itself.


wesm:
Resolved the stale-daemon compatibility finding in `1e10218`.

- Daemon runtime records now publish an explicit API protocol revision.
- `client.Ensure` requires version, protocol revision, and API key compatibility before returning any data client; a missing or mismatched revision takes the existing stop-and-replace path.
- Discovery used by `daemon status`/`stop` remains permissive so incompatible daemons can still be found and shut down.
- The real-binary lifecycle test removes the protocol revision from a live same-version daemon record and verifies that its PID is replaced before startup returns; unit tests cover both missing and mismatched revisions.

This prevents the new CLI from ever sending `run:false` to the legacy destructive handler through its normal `client.Ensure` path. Full Linux, Windows, and lint CI is green, and post-commit roborev job 1457 passed with no findings.

PR #5, opened 2026-07-09
Add scripted release preparation

wesm:
Docbank now has a real release to maintain, so future tags need the same repeatable, reviewable preparation path used by msgvault instead of ad hoc tag commands and GitHub-generated prose. This adds that shared operating model while retaining docbank-specific guards against cutting a release from dirty, divergent, non-main, or malformed-tag state.

The changelog helper feeds only the bounded commit log and diff summary to Codex or Claude, excluding documentation-only and reviewer-configuration changes. The release helper presents those notes for human confirmation, then creates and pushes an annotated semver tag. The publication workflow consumes the annotation body as the GitHub release notes, with generated notes retained only for lightweight or empty tags.

The synchronization check explicitly refreshes the remote-tracking main ref before comparing it to HEAD; this closes the review finding that a named-branch fetch could leave the guard stale.

Validation:

- full `CGO_ENABLED=1 make test`, `make lint`, and strict `make docs-build`
- `bash -n` and `shellcheck` on both scripts
- `actionlint` and zero-finding `zizmor --pedantic`
- real Codex changelog generation over the v0.1 commit range
- invalid version, non-main branch, unknown agent, and empty-output failure paths
- annotated-tag type/body extraction against live `v0.1.0`
- explicit origin/main refspec verified against `git ls-remote`

The initial `v0.1.0` tag was intentionally cut manually from merged commit `5046ff7`; its release workflow and all downloaded checksums passed, the macOS binary reports the expected version/commit, and authenticated `docbank update --check` reports already up to date.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -98,7 +98,7 @@ jobs:
           else
             : > release-notes.md
           fi
-          if [[ -s release-notes.md ]]; then
+          if grep -q '[^[:space:]]' release-notes.md; then
             echo "has_body=true" >> "${GITHUB_OUTPUT}"
           else
             echo "has_body=false" >> "${GITHUB_OUTPUT}"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,6 +25,7 @@ jobs:
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
+          fetch-tags: true
           persist-credentials: false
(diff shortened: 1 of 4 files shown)


wesm:
Resolved in `e079d4b`. The workflow now uses `grep -q '[^[:space:]]' release-notes.md`, so formatting newlines and whitespace do not count as a tag body.\n\nValidated the exact branch behavior against three local refs: a subject-only annotated tag falls back to generated notes, a lightweight tag falls back, and the populated `v0.1.0` annotation is detected. `actionlint`, zero-finding `zizmor --pedantic`, `prek`, and post-commit roborev job 1473 all pass.

PR #16, opened 2026-07-12
Expose all-loose storage recovery

wesm:
Packed storage needs a supported exit path as well as packing and compaction. Returning to canonical loose files is useful for recovery, direct inspection, or a deliberate storage-policy change, but the transition is unsafe if catalog authority moves before every live blob has been durably materialized and verified.

This PR exposes Kit’s existing all-pack preflight and unpack lifecycle through the authenticated daemon. Kit validates every source pack and footer before the first loose write, materializes and verifies all live content, atomically clears packed catalog authority, and only then retires source packs after active readers release them. Docbank contributes the API maintenance gate and product reporting; it does not recreate the physical transition.

Unpack changes representation only. Node IDs, paths, blob hashes, and readable content remain stable. The report names raw bytes restored and packs retired without presenting that value as net disk growth, since the operation may temporarily hold both representations. Repeated runs converge to zero work.

This completes the explicit status, pack, repack, and unpack operator lifecycle on the Kit packstore foundation.

### DIFF
--- a/cmd/docbank/cli_test.go
+++ b/cmd/docbank/cli_test.go
@@ -428,6 +428,32 @@ func TestDeletionHelpSeparatesTrashGCAndRepack(t *testing.T) {
 	assert.Contains(t, out, "requires a separate storage repack")
 }
 
+func TestStorageUnpackJSON(t *testing.T) {
+	_ = setupVaultHome(t)
+	for name, content := range map[string]string{"a.txt": "alpha", "b.txt": "bravo"} {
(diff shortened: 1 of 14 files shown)


wesm:
Closing by product decision: Kit's unpack primitive remains useful for tests, migrations, and emergency tooling, but docbank will not expose it as an ordinary API/CLI operation. Packed storage is the intended managed representation; public unpack would recreate small-file costs, require transient duplicate disk space, and leak physical-format management into the operator surface.

PR #35, opened 2026-07-13
Publish six-platform release artifacts

wesm:
Docbank now supports the full vault lifecycle on Linux, macOS, and Windows, but its release pipeline still publishes only three of the six supported OS/architecture combinations. That makes Windows support and macOS on Intel source-build-only despite the runtime contract now being complete.

This makes distribution match the product boundary:

- build Linux, macOS, and Windows archives for amd64 and arm64, using native Linux/Windows runners and CGO-enabled macOS cross-compilation;
- block release publication unless the exact six artifacts contain the expected executable, target architecture, embedded release version, and generated checksum entry;
- add shell and PowerShell installers that refuse missing, duplicate, malformed, or mismatched checksums and reject unexpected archive contents before publishing a binary; and
- document the supported installer, manual archive, destination, version-selection, and PATH behavior without adding Nix packaging.

The normal platform suites execute the actual shell installer on Unix and the actual PowerShell installer on Windows against generated release fixtures. They prove that a verified archive replaces the binary while checksum mismatches, duplicate entries, and unexpected payloads fail without changing an existing installation. The native runners also exposed and removed an optional Windows cmdlet dependency, leaving the PowerShell 5/7 path on framework cryptography primitives.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,9 @@ jobs:
           go-version-file: go.mod
           cache: true
 
+      - name: Check Unix installer
+        run: sh -n scripts/install.sh
+
       - name: Vet
         run: go vet -tags fts5 ./...
 
@@ -74,6 +77,13 @@ jobs:
         with:
           go-version-file: go.mod
           cache: true
+
+      - name: Check PowerShell installer
+        if: matrix.goarch == 'amd64'
+        shell: pwsh
+        run: |
+          $ErrorActionPreference = 'Stop'
+          [scriptblock]::Create((Get-Content -Raw scripts/install.ps1)) | Out-Null
 
       - name: Set up ARM64 C compiler
         if: matrix.goarch == 'arm64'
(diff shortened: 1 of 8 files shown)


wesm:
Addressed in `955dda3`: the entire `build-linux` container job now declares `defaults.run.shell: bash`, so both existing Bash-dependent steps and future run steps use the intended shell. Actionlint accepts the corrected workflow.

wesm:
Addressed in `6eb1769`:

- the write-capable publish job no longer executes any downloaded binary; native smoke tests remain in the read-only build jobs that produce the artifacts; and
- README/setup now state that v0.1.0 contains only Linux amd64/arm64 and macOS arm64, directing Windows and macOS amd64 users to source builds until the first six-platform release is published.

wesm:
Addressed in `61b18a4`. I checked the current msgvault workflow as requested: its native Ubuntu 22.04 container layout is the relevant prior art, but it carries the same unchecked direct Go archive download. Docbank now retains that native container/runner structure while using the same commit-pinned `actions/setup-go` invocation already used by msgvault's macOS/Windows jobs and by all Docbank CI/release builders. The direct `go.dev` download, extraction, PATH override, and redundant architecture field are gone.

PR #44, opened 2026-07-14
Keep pre-release metadata changes direct

wesm:
Docbank has no deployed storage compatibility boundary. Before the first public release, schema and deterministic JSONL changes should replace the development shape directly while remaining format v1; stale developer vaults are disposable.

This corrects the living architecture and agent contract accordingly. It removes the planned v2 bootstrap, logical cutover, downgrade-fence, old-binary, and legacy-conversion machinery while retaining the stable identity and audited-history intent. Audit activation is one guarded SQLite transaction, not a store-generation transition.

It also records an implementation rule for the load-bearing system we intend to build: metadata work must land as vertical changes to the live store, ingest, reachability, and backup/restore paths. Parallel schemas or codecs that production does not consume are out of scope; shared helpers are extracted only when those live paths use them.

Kata: `z7bc`.


wesm:
Superseded by #45, which now targets main and contains this metadata-v1 policy together with its first live ingest, backup, retrieval, and verification implementation. The review feedback on unconditional v1 import and chronology rules is fixed there in c4bdebc.

PR #69, opened 2026-07-17
Record inherited audited node creation

wesm:
An audited directory cannot safely accept new children unless membership inheritance and history recording happen in the same transaction. Otherwise creation is either permanently unavailable after enrollment or becomes a retention escape that leaves the new node purgeable.

This adds the smallest direct-creation slice: directory and file creation beneath an audited member now bind the child, its initial content when applicable, the parent revision change, topology transition, inherited baseline, scope chain, and allocator lineage as one Go-owned mutation. JSONL import and verification independently replay that mutation and reconcile the resulting projections instead of trusting the writer’s stored records.

The boundary is intentional. This covers Store.Mkdir, Store.CreateFile, and the embedded Put path. Filesystem ingest remains fail-closed because provenance is attached metadata and must be recorded atomically in its own focused follow-up; this PR does not add a partial ingest contract or speculative public audit surface.

Rollback, inherited directory/file creation, deterministic metadata round-trip, and adversarial membership-retarget coverage exercise the new authority path, and the repository’s CGO, pure-Go, race, lint, vet, and documentation gates are green.

Kata: qd6q.

### DIFF
--- a/internal/store/audit_baseline_projection.go
+++ b/internal/store/audit_baseline_projection.go
@@ -324,7 +324,7 @@ func initialBaselineTopology(
 			return nil, nil, err
 		}
 		parents[id] = parent
-		originValue, err := auditField(record, "origin")
+		originValue, err := auditField(record, auditOriginField)
 		if err != nil {
 			return nil, nil, err
 		}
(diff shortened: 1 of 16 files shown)


wesm:
Looks good here.

PR #71, opened 2026-07-17
Record audited filesystem ingest

wesm:
Audited vaults can record direct child creation, but their primary filesystem import path still had to fail closed because ingest identity, file creation, and provenance were separate metadata mutations. Accepting that split would make protected history unable to prove the source fact that introduced a document.

This keeps an ingest run authority-free until a file actually imports, then commits the file, initial version, sticky membership, immutable ingest/provenance facts, and canonical attached-metadata delta as one Go-owned transaction. Independent replay reconstructs the attached facts during JSONL import and verification rather than trusting the writer’s projections. Idempotent repeats remain non-mutating, and a failed audit advance rolls back the file and its source metadata together.

The change extends the existing audited node-creation writer instead of introducing a parallel record builder. It does not expose unfinished audit controls or advertise internal capability in the user documentation. The CGO and pure-Go suites, focused race coverage, lint, and strict documentation build are green.

Kata: 9j5e.

### DIFF
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -401,7 +401,7 @@ func (ing *Ingester) AddPathsWithOptions(
 func (ing *Ingester) addTree(
 	ctx context.Context,
 	rep *Report,
-	ingestID string,
+	ingestRun store.IngestRun,
 	destDirID int64,
 	sourceRoot, walkRoot string,
 	excludes exclusions,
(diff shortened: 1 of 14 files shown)


wesm:
Addressed in 097b762. Replay initialization now builds the genesis ingest set before accepting provenance and rejects any provenance whose ingest appears only later. The regression test constructs that exact temporal mismatch from a real later audited-ingest delta.

PR #102, opened 2026-07-20
Build disposable embedding generations

wesm:
Docbank can now prepare its verified current text for semantic search without making model output part of the archive. After configuring an OpenAI-compatible encoder, a person or agent can run `docbank embeddings build`, watch progress, and inspect retained generations with `docbank embeddings list`. Byte-identical documents are encoded once.

The disclosure boundary is explicit: configuration alone sends nothing, and a requested build sends only successfully extracted text from current live documents. It does not send paths, tags, provenance, audit records, historical versions, trashed content, or original binary files.

Vectors live in a disposable local `vectors.db`, outside JSONL and backup authority. Builds resume incomplete work, and a replacement model becomes active only after it covers the complete current mirror; a failed replacement leaves the prior complete generation active. Restoring a vault therefore restores documents and metadata, then the vector index can be rebuilt.

This also preserves the embedding boundary for Go applications. The public `go.kenn.io/docbank` package does not import the daemon vector layer, so embedders that do not need semantic indexing can continue using pure-Go SQLite without a CGo vector dependency. The official daemon owns the optional vector backend.

This PR deliberately stops at generation construction and inventory. Semantic and hybrid query will follow as a separate, unstacked change built on this substrate.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,7 +38,8 @@ jobs:
           DEBIAN_FRONTEND: noninteractive
         run: |
           apt-get update
-          apt-get install -y --no-install-recommends build-essential ca-certificates file git gzip tar
+          apt-get install -y --no-install-recommends \
+            build-essential ca-certificates file git gzip libsqlite3-dev tar
 
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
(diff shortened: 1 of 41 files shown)


wesm:
This project will need some work, since some documents may benefit from a distillation pass before being embedded.

wesm:
We are deliberately tabling embeddings until the document-distillation and evaluation contract is clearer. Keeping this implementation open would imply a product direction we have not validated; the deferred kata issues preserve the research context for a future restart.

PR #104, opened 2026-07-21
Compress worthwhile loose document storage

wesm:
Documents can spend a long time in loose storage before packing, and large objects above Docbank’s current 64 MiB packing policy stay loose indefinitely. Compressible content should not occupy its full decoded size during either period.

This upgrades Docbank to Kit v0.11.0 and enables managed zstd for new loose objects of at least 4 KiB when compression saves at least 10%. A 100 MiB text archive can therefore occupy far less physical space while keeping the exact same SHA-256, logical size, stable document identity, and API bytes. Tiny or incompressible objects remain raw. Existing raw vaults continue working as-is; startup does not rewrite them.

The physical encoding remains an implementation detail. Reads verify decoded bytes, packing accepts either loose representation, and JSONL backup/restore preserves logical content rather than the `.zst` filename. Storage status and GC deliberately report physical loose bytes, including both copies if repair evidence ever leaves raw and compressed representations together.

The tradeoff is temporary capacity during eligible writes: Kit privately stages the complete raw object and its zstd candidate before choosing what to publish, so concurrent large writes need room for both candidates. Cancellation or failure removes that private staging without granting document authority.

The round-trip coverage includes compressed write and read, raw fallback, physical accounting and deletion, compressed-loose packing and sparse repack, and verified JSONL backup into a restored packed vault across both SQLite implementations.


wesm:
Superseded by #103, which now contains the useful parts of this PR: the standalone 4 KiB / 10% managed zstd policy, human and architecture documentation, raw-plus-zstd physical accounting, and verified backup/restore coverage. #103 also supplies the more complete indexed physical-authority model and embedded create/repair APIs, so keeping two implementations would duplicate and weaken the storage boundary.

PR #134, opened 2026-07-23
Publish complete release history and web UI guide

wesm:
Docbank.ai will present a complete, dated history from v0.1.0 through v0.11.0, with the changelog immediately beside the overview and the current web application documented as a real browsing workflow rather than a source-only preview. Users reading any current guide will no longer encounter “next release” or “newer than” fences for commands that have already shipped.

That consistency matters because the site is published as a release snapshot, not continuously from `main`. Mixing development-era availability notes into that snapshot made shipped features appear unavailable and left architecture pages disagreeing about multi-scope audited backup. The repository guidance now makes the release-snapshot boundary explicit for future documentation work.

The web guide includes a vault browser and extracted-text search capture generated from the exact v0.11.0 frontend and pure-Go backend in Docker against a synthetic vault. The PNGs live on the orphan `docs-assets` branch, with capture provenance, so the guide can remain visual without adding binary weight to the main source history.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -82,17 +82,20 @@ Instructions for autonomous coding agents working in this repository.
   a general API or CLI command; reserve it for tests, migrations, or a
   purpose-built emergency recovery workflow with a demonstrated need.
 - Documentation is not the implementation tracker. User- and agent-facing
-  pages describe shipped capabilities and current limitations. Architecture
-  and internal pages may preserve durable future design under explicit
-  `!!! info "Planned"` admonitions, but must not carry task breakdowns,
-  sequencing, ownership, or completion criteria. `docs/roadmap.md` is the one
-  high-level public product-status view; kata is the sole source of truth for
-  actionable work and its status.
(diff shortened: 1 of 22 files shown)


wesm:
Addressed the ambiguity in `7770ccd`. The review premise that documentation deploys continuously from `main` does not apply to this repository: the site is published only from release tags. `AGENTS.md` now distinguishes the two states explicitly—`main` carries the coherent documentation candidate for the next tag, while release preparation must verify every documented capability is present in that tag and defer anything that will not ship. Merged capabilities therefore remain documented directly in present tense without user-facing source-build or “next release” fences.

wesm:
Clarified in `534aabd`. There are no rendered previews, deployments, or public documentation builds of `main`; the release tag is the first publication boundary. Homepage and setup files on `main` are source inputs for the next tagged release, not an independently published surface. Requiring release qualifiers there would encode a publication mode that does not exist and would leave stale timing notes to remove at every tag. Release preparation instead verifies that every documented capability is present in the tag and defers anything that will not ship.

PR #168, opened 2026-08-16
Add provider-neutral document normalization

wesm:
Docbank now provides deterministic document normalization as a normal public Go package. Applications can turn source units into stable text, headings, chunks, spans, and checksums without opening a vault or using Docbank's daemon, storage, queue, or application code.

The version-2 policy has one caller choice: the total document character limit. Chunk sizes and the other structural values are fixed because changing them would change document identity. Expected results come from the pinned Msgvault PR #616 source with the approved soft-break separator correction applied before generation. This keeps the oracle independent while avoiding a known defect that joined words across ordinary Markdown line breaks.

This pull request deliberately stops at the provider-neutral boundary. It does not add Mistral uploads, staging, capability probes, or upload authorization; those safety-sensitive pieces will arrive together in a separate pull request.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# The compatibility bundle is a byte-for-byte oracle, not a text file.
+document/internal/compattest/testdata/document-compat-v1.json -text
--- a/docs/superpowers/specs/2026-08-16-document-understanding-implementation-design.md
+++ b/docs/superpowers/specs/2026-08-16-document-understanding-implementation-design.md
@@ -1,106 +1,37 @@
-# Document-understanding implementation delivery
-
-**Status:** Approved design
+# Document-understanding package architecture
 
 **Date:** 2026-08-16
 
-**Implementation work:** `e9jz`
+## Purpose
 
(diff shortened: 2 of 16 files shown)


wesm:
cc @salmonumbrella

wesm:
Both requested problems are fixed.

- `087282c` preserves a separating space across ordinary Markdown soft breaks and adds a focused regression test. Version 2 changed in place because this package and the source Msgvault work are both still unmerged and have no consumers.
- `0fda9e6` marks the raw compatibility bundle as binary so Windows cannot rewrite its line endings. A simulated `core.autocrlf=true` checkout now produces the pinned SHA-256.

The corrected oracle was regenerated twice from the pinned Msgvault source plus the recorded soft-break patch. The new raw digest is `2f968cce51b67ff5969bab6e5502d5792f815d7cd6c416802ee16ba0f8e36b89`. Both complete SQLite-mode suites, lint, strict docs, and hooks pass.

salmonumbrella:
watching

PR #172, opened 2026-08-18
Add media detection and fail-closed Voyage multimodal embedding contracts

wesm:
Applications can now detect images and video and embed them through Voyage AI using the same fail-closed contract Docbank already provides for Mistral OCR: format detection and bounds live in one place, and nothing is uploaded until an operator has run an authenticated capability probe and recorded which formats, animated images, video, query modes, and batch shapes the provider actually handles.

Two new packages, mirroring `document` and `document/mistral`:

- `document/media` sniffs JPEG, PNG, WebP, GIF, and MP4 containers, reads only dimensions, frame count (including animated WebP and APNG), and declared duration — never pixels — and evaluates results against byte, pixel, duration, and still/animated/video bounds with stable machine-readable reasons. Under a duration cap, video whose duration cannot be measured is refused. `mediatest` builds synthetic containers for consumer tests.
- `document/voyage` pins the Voyage multimodal model, dimension, and endpoint; runs a deterministic fixture probe (JPEG, PNG, and GIF generated; WebP and MP4 from operator-supplied synthetic seeds); validates and strictly decodes the resulting manifest; and derives per-capability upload authority through `Policy.Authorize`. The client accepts only the probed input shapes, re-detects every media part and serializes from the detected metadata so caller metadata cannot misdescribe bytes, authorizes image queries per format and text-then-image queries as their own probed shape, decodes responses strictly (unique index per vector, no duplicate keys), refuses batches over the policy's item and byte limits before allocation, never follows redirects, classifies provider failures without echoing response bodies, and honors `Retry-After`. `Fingerprint` gives applications a policy identity to bind consent records to. `voyagetest` supplies synthetic manifests, and a `voyage_probe` build tag runs the live probe.

The boundary is unchanged: Docbank owns detection, bounds, transport, capability evidence, and request identity; the importing application keeps credentials, human consent, budgets, orchestration, vector storage, and search. Docbank does not compute, store, or serve embeddings for its own vault, and neither package imports the vault, database, daemon, or queue.

The strict manifest JSON checks the Mistral package used are shared through `document/internal/manifestjson` so both providers validate manifests the same way.


wesm:
cc @salmonumbrella

salmonumbrella:
watching

salmonumbrella:
@wesm lfg!

wesm:
Merged! I'll work on rebasing / refactoring the msgvault PR

PR #174, opened 2026-08-20
Probe interleaved documents per format and mixed batches

wesm:
Applications embedding message or document context alongside media can now do so for every supported format, not only PNG. The previous contract authorized text-then-media documents solely with PNG media and batches solely as media-only PNG, so a JPEG, WebP, GIF, or MP4 attachment could never carry its owning context to the provider.

Each format now has its own interleaved capability (`interleaved_text_jpeg` through `interleaved_text_mp4`), probed by demonstrating that both parts contribute: swapping the media between contrasting fixtures of the same format and animation state moves the vector, and swapping the text moves it again — cross-format comparisons are never used, so container-only sensitivity cannot pass as pixel contribution. Every format has a contrasting variant fixture: generated for JPEG and GIF, the red PNG for PNG, and two additional operator seeds (`image_webp_alt.webp`, `video_mp4_alt.mp4`) for the formats the standard library cannot encode; validation rejects a variant whose bytes match its primary.

The batch probe places every passed format's primary AND contrasting variant, plus both of its independently proven text composites, into batches packed against the policy's item and estimated request-byte limits before any body is built. A boundary batch cycles members up to the largest permissible size so the recorded item limit is exercised even when it exceeds the member pool, and one additional batch must actually hold several formats — the capability fails when no permissible request can mix them. Each member must match an independently embedded reference of its identical input and beat a planted contrast partner — the same-format sibling for media members, the other-text composite for composites — so same-format reordering, duplication, and text dropping inside batches are all observable; the paired composites make text dropping contradict at least one inequality regardless of media colinearity. Cross-format inequalities are never asserted, because different containers of the same picture may legitimately embed alike. A policy whose limits cannot fit any multi-item request records that multi-item batches have no evidence instead of passing vacuously, and when nothing passed a single reference request still runs so the result records the provider's real failure mode without ever counting as batch evidence.

`EmbedDocuments` requires the format's own interleaved capability for `[text, media]` and the batch capability for more than one input; each member is still authorized individually. Combined text-and-image queries remain PNG-only. The manifest schema version advances to 2 and the fixture contract to 3 because the capability list, fixtures, and evidence checks changed; existing manifests must be re-probed.


wesm:
One reviewed finding is deliberately not implemented: requiring the probe to reach the configured item limit when the request-byte cap prevents it (or recording only the largest observed count). Production requests are bounded by the same MaxRequestBytes check the probe packs against, so any batch the probe could not send, a consumer cannot send either. The residual gap — production media small enough to reach higher item counts than the probe's fixtures allowed — is a fixture-size delta, and closing it would change manifest semantics: Authorize treats a smaller recorded MaxBatchItems as a target mismatch, not a cap, so recording the observed size would need a schema and consumer contract change disproportionate to the risk. With realistic operator seeds the boundary batch reaches the configured limit outright; the first consumer batches at two.

wesm:
Extending the note above to cover its per-format variant, which was also reviewed: a format whose fixtures are too large to share any permissible multi-item request is likewise left with singleton-only batch evidence by design. Both are the same gap — probe fixtures larger than production media — and both are bounded by the identical MaxRequestBytes check on the production path. The review loop on this probe is closed as of eba9796; further evidence permutations should be weighed against the live-probe gate, since every capability here remains off until an operator's authenticated probe passes.

Issue #176, opened 2026-08-22
Roadmap: provider-neutral document processing and semantic retrieval


salmonumbrella:
## Design

Docbank owns the durable sanitized Markdown, normalized evidence, embedding sets, rebuildable indexes, and document-candidate retrieval. Msgvault owns the attachment/message mapping, source visibility, mirroring, source-version fence, and its app surfaces.

### Foundation

- [x] F1 — #180 — canonical `source-evidence/v1` and
  `normalized-evidence/v1` codec
- [x] F2 — #182 — self-describing safe Markdown rendition, bounded navigation,
  and model-independent lexical evidence
- [x] F3 — #183 — processing profiles and layered fingerprints
- [x] F4 — #184 — immutable derivative catalog and deterministic metadata JSONL
- [x] F5 — #185 — derivative blob and lexical-generation publication
- [x] F6 — #186 — exact legacy plain-text authority cutover
- [x] F7 — #187 — derivative backup and provider-free restore
- [x] F8 — #188 — complete-manifest purge and lease-aware derivative GC
- [x] F9 — #194 — auxiliary MD5 fingerprints for original document bytes and retained sanitized `.md` artifacts; SHA-256 remains authoritative
- [x] F10 — #195 — local original-file metadata extraction with normalized common fields, bounded format-specific evidence, privacy-safe disclosure, and deterministic backfill

### Rendition runtime

- [x] #181 — provider-neutral OCR contract, hosted Mistral bridge, and local
  GLM-OCR deployment
- [x] R1 — #189 — enforceable rendition-provider contract
- [x] R2 — #190 — exact-byte inspection and sealed authorized upload
- [x] R3 — #191 — DNS-to-dial provider egress
- [x] R4 — #192 — append-only processing consent and restore incarnation
- [x] R5 — #193 — standard `docbank-rendition/v1` bridge
- [x] R6 — #196 — verified in-process plain-text provider
- [x] R7 — #198 — Docling rendition adapter
- [x] R8 — #199 — Datalab rendition adapter
- [x] R9 — #201 — Mistral rendition adapter, after the relevant #175 bounds
- [x] R10 — #202 — resumable rendition jobs and fenced publication
- [x] R11 — #203 — self-hosted Marker rendition adapter
- [x] R12 — #204 — verified local-process PyMuPDF PDF-text provider
- [x] R13 — #205 — hosted LlamaParse rendition adapter
- [x] R14 — #206 — hosted Reducto rendition adapter
[...]

PR #180, opened 2026-08-23
Add canonical provider-neutral evidence contracts

salmonumbrella:
## What changed

Docbank now has versioned provider-neutral `source-evidence/v1` and `normalized-evidence/v1` contracts for ordered units, locators, hierarchy, geometry, tables, retained artifacts, confidence, and explicit omissions.

Normalization produces canonical UTF-8 NFC with LF endings, fixed-point confidence, stable generation-local IDs, exact canonical JSON, and a SHA-256 checksum. Provider-local IDs never enter durable identity. Unknown families, invalid geometry or ranges, mutable artifact pointers, malformed completeness claims, and evidence outside the caller’s finite policy fail closed.

## Why

Provider adapters need one durable evidence language before rendition, search, and retrieval can build on their output. Without it, page, table, geometry, hierarchy, and omission evidence stay provider-specific and cannot be checksummed or compared consistently.

## Usage

Normalize provider evidence and produce its canonical bytes:

```go
policy, err := document.NewEvidencePolicy(maxDocumentCharacters)
if err != nil {
	return err
}

normalized, err := document.NormalizeEvidenceV1(sourceEvidence, policy)
if err != nil {
	return err
}

canonicalJSON, checksum, err := document.MarshalNormalizedEvidenceV1(normalized)
```

This adds the evidence contract only; it does not change storage or serving.

Part of #176 (F1).


wesm:
looking

wesm:
will merge pending green CI

PR #181, opened 2026-08-23
feat(document): add provider-neutral local OCR

wesm:
Docbank can now use hosted Mistral OCR or self-hosted GLM-OCR and return the same ordered, page-aware result. Existing Mistral safety behavior and `mistral-ocr-4-0` support remain intact.

## Provider boundary

- `document/ocr` defines the shared result and error contract.
- The GLM-OCR adapter accepts PDFs and page images, preserves blank pages, validates structured output, and does not retry invalid input.
- Fingerprints separate derived output when the provider, model, policy, or deployment changes.

## Local deployment

- The ARM64 deployment pins the image, models, SDK, dependencies, renderer, and pipeline configuration. The engine refuses health and OCR requests when the running deployment does not match.
- Only the page-aware adapter is exposed on loopback. This profile is for a dedicated or single-user host; shared hosts need authenticated local transport.

## Evidence and limits

- A synthetic DGX Spark-class GB10 baseline completed 8/8 cases and nine pages at 2.321 pages/second. It used about 12.6 GiB of GPU memory for inference and 1.3 GiB for layout processing.
- The measurements predate the renderer update. This PR makes no quality-equivalence claim against Mistral.
- PaddleOCR-VL-1.5 was not benchmarked because its pinned SM120 image was unavailable.

PR #177 builds retrieval and evaluation above these normalized documents. The branches do not overlap.


wesm:
cc @salmonumbrella

PR #182, opened 2026-08-23
Build safe Markdown renditions from canonical evidence

salmonumbrella:
## What changed

Docbank can now turn canonical evidence into sanitized Markdown, normalized units, and model-independent lexical segments in one bounded pass. The rendition keeps useful document structure—headings, lists, task checkboxes, tables, code, safe links, and page boundaries—while stripping active HTML, scripts, styles, unsafe URLs, and provider control markup.

Stable checksums bind the displayed Markdown and lexical index to the same source rendition. If a source or output budget is exhausted, Docbank emits a structurally valid prefix instead of malformed Markdown.

## Why

Provider output cannot become durable document authority as-is. We need one deterministic boundary that preserves useful structure without letting active content, unsafe links, or provider-specific markup leak into stored Markdown or search.

## Usage

Build a safe rendition from canonical evidence:

```go
policy, err := document.NewRenditionPolicy(normalizePolicy)
if err != nil {
	return err
}
rendition, err := document.BuildRenditionV1(evidence, policy)
```

This adds the document contract only; it does not change storage or serving yet.

Part of #176 (F2).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,3 +2,5 @@
 document/internal/compattest/testdata/document-compat-v1.json -text
 # The evidence golden bundle is a byte-for-byte oracle, not a text file.
 document/testdata/normalized-evidence-v1.golden.json -text
+# The rendition golden bundle is a byte-for-byte oracle, not a text file.
+document/testdata/rendition-v1.golden.md -text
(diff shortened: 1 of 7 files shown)


wesm:
Working on #180 now

wesm:
rebasing

wesm:
Does not reproduce

PR #183, opened 2026-08-23
Add stable processing profiles and fingerprints

salmonumbrella:
## What changed

Docbank now has canonical `processing-profile/v1` policy bytes and separate fingerprints for rendition, evidence, lexical work, embedding inputs, vector spaces, retention, and provider disclosure. Each fingerprint changes only when its own semantic inputs change, so switching an embedding model does not invalidate a rendition and changing retention does not rebuild evidence.

Named TOML profiles cover rendition, embedding, retrieval, and complete processing policy. Credential references stay symbolic as `credential:<name>`; runtime secrets never enter canonical profile bytes.

## Why

One whole-profile hash would rebuild unrelated work, while incomplete component hashes could reuse artifacts across materially different policy or disclosure boundaries. The catalog and workers need stable, narrow cache identities before they can safely share work.

## Usage

Resolve and fingerprint a named profile:

```go
profile, err := cfg.ProcessingProfile("archive")
canonical, fingerprints, err := document.CanonicalProfile(profile)
```

This PR defines and validates policy only. It does not execute providers or publish derivatives.

Part of #176 (F3). Stacks on #182.


wesm:
Working on #180 now

wesm:
rebasing

PR #184, opened 2026-08-23
Add the durable document derivative catalog

salmonumbrella:
## What changed

Docbank can now record immutable rendition builds, artifacts, source-version attachments, and exact serving heads in a durable derivative catalog. Imports publish only after every source and derivative byte matches its recorded hash and size; missing or corrupt data rolls the whole import back.

Processing metadata round-trips through deterministic JSONL in both SQLite implementations. Artifact policies explicitly define required, optional, and multi-valued roles, so an accepted build cannot silently omit or add outputs.

## Why

Renditions need one auditable source of authority before search, migration, backup, or garbage collection can use them. A partially imported or corrupt derivative must never become visible just because its catalog row exists.

## Usage

There is no new operator command in this PR. Later publication and worker paths use the catalog to stage builds, attach them to exact source versions, and move a verified head into serving authority.

Part of #176 (F4). Stacks on #183.


wesm:
rebasing

wesm:
False positive

PR #185, opened 2026-08-23
Publish renditions and lexical generations atomically

salmonumbrella:
## What changed

Docbank can now publish a verified rendition and its lexical search generation as one atomic authority. A failure after blob write, catalog staging, FTS construction, or immediately before the head flip leaves the previous publication serving.

Every retained payload is checked against its exact hash and size. Canonical evidence is rebuilt under the captured rendition policy, lexical manifests cover their exact content, and active attachments are selected in the same SQLite snapshot used to read their rows.

## Why

The derivative catalog from #184 was intentionally dormant. Publishing it piecemeal would expose half-built state, stale search rows, or shared builds under the wrong content version. The head should move only when the complete rendition and lexical generation agree.

## Usage

There is no standalone CLI yet. Processing workers can stage a rendition build and lexical generation, validate the complete manifest, then publish both together. Readers keep using the previous head until that transaction succeeds.

Part of #176 (F5). Stacks on #184.

### DIFF
--- a/internal/backupapp/app_test.go
+++ b/internal/backupapp/app_test.go
@@ -321,17 +321,18 @@ func TestJSONLSnapshotRestoresRenditionBytesBeforeVerifyingHeads(t *testing.T) {
 		},
 	}
 	require.NoError(t, fixture.metadata.StageRenditionBuild(t.Context(), build))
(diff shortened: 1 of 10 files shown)


wesm:
rebasing

PR #186, opened 2026-08-23
Move legacy plain-text renditions into canonical authority

salmonumbrella:
## What changed

Existing successful `plain-text` v1 records now migrate into canonical rendition builds, per-version attachments, rendition heads, and one complete lexical generation when a current or released vault opens. The cutover preserves the exact stored UTF-8 bytes and validates the replacement FTS generation against current name and text search before publishing it.

Failed, missing, invalid, or obsolete rows stay outside serving authority and remain queued for fresh extraction. The legacy cache and released source database remain recoverable but stop acting as a second live search authority.

## Why

Serving from both the released plain-text cache and the new derivative catalog would make upgrades and retries vulnerable to mixed or incomplete state. Existing vaults need one deterministic cutover without rewriting their extracted text or throwing away recovery data.

## Usage

No manual migration command is required. Opening an older supported vault performs the cutover automatically; the previous head keeps serving until the replacement catalog and lexical generation are complete.

Part of #176 (F6). Stacks on #185.

### DIFF
--- a/internal/store/extraction.go
+++ b/internal/store/extraction.go
@@ -313,6 +313,20 @@ func (s *Store) RecordExtraction(ctx context.Context, result ExtractionResult) e
 		if !exists {
 			return fmt.Errorf("blob %s: %w", result.BlobHash, ErrNotFound)
 		}
+		var storedVersion int64
+		err := tx.QueryRowContext(ctx, `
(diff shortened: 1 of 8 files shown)


wesm:
rebasing

PR #187, opened 2026-08-23
Back up and restore document derivatives without providers

salmonumbrella:
## What changed

Backups now include the complete catalog-authorized document authority: current source versions, rendition artifacts, and staged rendition-build sources. Snapshot metadata says which derivative classes are included, locally reconstructible, or provider-dependent, with deterministic counts, bytes, blob totals, and checksums.

Restore validates the logical and physical authority after materializing loose or packed bytes, then rebuilds the lexical projection locally before any restored head can publish. Missing, corrupt, or inconsistent data leaves an existing target unchanged.

## Why

The catalog was durable inside one vault, but backup still treated its bytes as generic storage. That could retain sensitive provider output without disclosing it or restore a vault whose catalog, blobs, and search head did not agree.

## Usage

Use the existing backup and restore commands. Derivative authority is included and reported automatically, and restore never calls a rendition or embedding provider.

Part of #176 (F7). Stacks on #186.

### DIFF
--- a/internal/backupapp/app.go
+++ b/internal/backupapp/app.go
@@ -21,18 +21,19 @@ import (
 // snapshot. Physical pack rows are deliberately excluded: restore may choose a
 // different loose/packed representation while preserving the same archive.
 type Stats struct {
-	Nodes           int64 `json:"nodes"`
-	Files           int64 `json:"files"`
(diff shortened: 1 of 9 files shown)


wesm:
Rebasing

PR #188, opened 2026-08-24
Purge document derivatives without touching originals

salmonumbrella:
## What changed

Docbank can now purge every live derivative for a document while preserving the original file, its history, and immutable backup copies. One audited transaction revokes attachments and heads, removes lexical segments and search rows, drops generation manifests, and records a durable suppression so a restart cannot silently recreate the same work.

Active readers and workers keep their exact generations, shared builds survive while any attachment or root still needs them, and in-flight publication cannot race purge. Exact packed-byte receipts survive interruption and force primary packs to be rewritten even when they are otherwise too dense for ordinary repack. If a derivative also exists in a packed secondary store that Docbank cannot rewrite, purge reports the incomplete erasure instead of claiming success.

## Why

Deleting a Markdown blob is not a real purge if search rows, catalog heads, shared builds, or packed bytes still retain the derivative—or if background work immediately recreates it. Purge needs to remove the complete live authority and make that decision durable without touching source content.

## Compatibility

Storage remains at schema version 4 and metadata JSONL remains at format version 1. Vaults from v0.14.0 migrate directly to the combined current layout; no intermediate unreleased schema or compatibility layer is required.

## Usage

Request a derivative purge for the target document. For example, purging `contract.pdf` removes its sanitized Markdown and derived search text but leaves the original and its history intact. A later rebuild requires explicit authorization.

Backup repository history is reported separately and is never rewritten by a live-vault purge.

Part of #176 (F8).


wesm:
rebasing

PR #189, opened 2026-08-24
Define the enforceable rendition provider contract

salmonumbrella:
## What changed

Docbank now has one provider-neutral contract for rendition descriptors, one-shot uploads, authorization, results, receipts, and classified errors. The boundary rejects stale or mismatched authority, changed provider or upload metadata, unsupported formats, oversized output, invalid evidence, and unsanitized provider failures.

Credentials, raw response bodies, storage, jobs, and consent stay outside the provider contract.

## Why

Every adapter needs the same enforceable boundary before it can receive document bytes. Otherwise provider-specific behavior can leak into durable authority or escape the exact source, policy, and output limits Docbank approved.

## Usage

Provider implementations expose a descriptor and accept only an authorized one-shot upload plus matching rendition authority. Callers validate the result through the shared contract before anything can enter the derivative catalog.

There is no standalone operator command in this PR; concrete adapters build on this boundary.

Part of #176 (R1). Stacks on #188.

### DIFF
--- a/document/doc.go
+++ b/document/doc.go
@@ -1,6 +1,10 @@
-// Package document converts provider-neutral source evidence into deterministic
-// normalized units, headings, spans, and chunks.
+// Package document defines the storage-neutral rendition-provider boundary and
+// converts provider-neutral source evidence into deterministic normalized
+// units, headings, spans, and chunks.
 //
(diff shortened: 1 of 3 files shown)


wesm:
rebasing

PR #190, opened 2026-08-24
Bind provider uploads to the exact inspected bytes

salmonumbrella:
## What changed

Docbank now inspects supported text, document-container, PDF, image, video, and WAV inputs under finite local bounds, then seals the exact bytes, detected format, provider descriptor, processing profile, and disclosure policy into one portable capability record.

The authorized upload copies and syncs the source, reopens it through descriptor-relative no-follow APIs, verifies file identity and a second full SHA-256, reinspects the held bytes, and removes the pathname before an adapter can read. Deceptive, encrypted, nested, externally referential, malformed, substituted, or locally unbounded inputs fail closed.

## Why

A successful preflight means nothing if the provider can receive different bytes later. The inspected source and adapter-visible source need to be the same held, hash-verified object, with no vault path or reopen capability crossing the boundary.

## Usage

Provider orchestration first inspects a source into a capability record, then creates a one-shot authorized upload from that same file authority. Deserialized records can describe prior inspection but cannot recreate local upload authority.

There is no standalone CLI in this PR.

Part of #176 (R2). Stacks on #189.


wesm:
rebasing

wesm:
## Inverted the resource-attribute check

Two of the three findings were already implemented in `1c4a076`, the commit reviewed. Verified by probe: an EPUB pointing `container.xml` at `OPS/package.dat` with an external `href` is rejected, and `noNamespaceSchemaLocation` is rejected.

The third was real — `<style>` CSS, `poster`, `object[data]`, `form[action]` passed through.

Rather than add those names, every attribute is now checked, with a small exemption list for values that name a vocabulary rather than a resource (namespace declarations, `xml:base`, OOXML `Type`, `dcterms` `property`). That list is bounded by the specs; the resource-attribute list isn't. This also catches `xlink:href`, `embed src`, and `background`, which nobody had enumerated. `<style>` element CSS is parsed with the tokenizer already used for the `style` attribute.

**Behavior change worth a look:** "external" now means *names a host* (or a `file:`/`data:`/`javascript:`-style scheme), not *parses as absolute*. The old test rejected real LibreOffice files on `ref="A1:C3"` and `config:name="ooo:view-settings"`.

Verified against real LibreOffice `.epub`/`.xlsx`/`.ods`; new tests mutation-checked; suite and lint green.

Documented on `inspectXML` as best-effort, not a boundary — scripted and entity-split references stay out of reach (`<meta http-equiv="refresh">` still passes). The #189 provider contract is the boundary.

wesm:
Fixed the first finding — it was a regression I introduced in `27b1b46`, now `cc1ed46`.

Switching "external" to mean *names a host* fixed the false positives on real spreadsheets (`A1:C3`, `ooo:view-settings`) but let hostless network references through. Confirmed all three forms reached `eligible`:

```
http:169.254.169.254/latest/meta-data   eligible
https:/tracker.example                  eligible
http:/\\169.254.169.254\latest          eligible
```

Now the scheme decides when no host was parsed: a scheme a consumer resolves on its own (`http`, `https`, `ftp`, `file`, `data`, …) is external regardless, while a colon-bearing identifier like `A1:C3` or `urn:uuid:` stays local. All three forms now rejected, real LibreOffice `.epub`/`.xlsx`/`.ods` still clean, regression tests added and mutation-checked.

**Second finding — not taking it without a decision from a human.** Two reasons:

1. The escalation isn't established. An embedded PDF's bytes are already counted against `MaxEntryBytes` and `MaxExpandedBytes`, so "unmeasured downstream processing" needs a concrete path where a provider recurses into a non-ZIP resource. The ZIP-signature check exists because a nested archive's expansion escapes the outer entry sizes; that argument hasn't been made for the other formats.

2. The obvious fix looks likely to reject real files. Embedded OLE objects in XLSX and PPTX are compound documents stored as `xl/embeddings/oleObject1.bin`; `.bin` isn't in `looksNestedContainer`, so they pass today and would start failing if compound-file magic were matched by signature. That's the same shape as the false positives I just spent two rounds fixing.

Happy to add the signature checks if you want them — `%PDF-` and the compound-file magic are both already in `formatdetect`, so it's small. But it's a deliberate tradeoff against real-world files, not a free hardening, so I'd rather you call it.

wesm:
Fixed the High finding in `3f37544`, and filed the PDF one as follow-up rather than fixing it here.

**EPUB resource selection.** Confirmed the bypass first: the same chapter renamed `.xhtml` → `.dat`, declared as XHTML in the manifest, went from `external_reference` to `eligible`.

This was the third instance of one bug — inspection picked entries by filename suffix. The container `rootfile` fix covered the package document; chapters, SVG, and stylesheets were still reachable under any name. So rather than extend the suffix list a third time, inspection now reads the manifest and follows the declared media type, which is how a reader decides what a resource is. Suffix matching stays alongside it, so an entry absent from the manifest is still covered by its name — neither signal alone covers both cases.

Closed and tested: neutral-named XHTML, SVG, and CSS. Real LibreOffice `.epub`/`.xlsx`/`.ods` still eligible; regression tests mutation-checked; suite and lint green.

**PDF `SubmitForm`/`ImportData` — not fixing in this PR.** It extends a classifier that already exists, so it's small, but dereferencing a form action needs a viewer that submits forms, which isn't what happens to these bytes downstream. Better as its own change against a stated threat model than as another round here.

Also still open from earlier and worth its own issue: `document/internal/formatdetect` has no test files, and it holds the PDF parsing.

wesm:
Fixed the manifest finding in `595512b`. Not taking the other two — reasons below.

**Manifest href with query/fragment** — real, and a regression from `3f37544`. Confirmed: `href="chapter.dat"` rejected, `href="chapter.dat?v=1"` eligible. Path resolution rejected query/fragment forms and the item was then dropped from the declared-type map, so the entry fell back to filename matching. A query selects a position inside a resource, not a different resource, so it is now stripped for the archive lookup and the declaration is kept. Regression test added and mutation-checked.

**Unbounded read before size rejection** — declining. `MaxSourceBytes` is the policy's own declared ceiling, so reading up to it is the bound working, not a bypass of it. Tightening the limit to `ExpectedBytes+1` is not free either: it would collapse the `source_bytes_exceeded` sealed record into a plain mismatch error, since you cannot tell "larger than the ceiling" from "not the declared size" without reading past the declared size. Worth doing only if that record state is expendable.

**ZIP entry-count checked after `zip.NewReader`** — declining. The central directory is bounded by the source, which is already capped at `MaxSourceBytes` before parsing, so the work is bounded. Enforcing during parse means not using `archive/zip`, which is a large change for a bound that already holds.

Both are reasonable as follow-up issues against a stated threat model. I would rather not keep reshaping this PR — the last three rounds have found regressions introduced by the previous round's fix rather than defects in the original change, and nothing imports `document/upload` outside its own tests yet.

wesm:
Fixed the cancellation handoff and added two review boundaries to `.roborev.toml` — `1e29e0c`. Not taking the rest.

**Cancellation handoff.** Real. Cancellation can land after `Authorize` checks the context but before the caller receives the upload; the callback then closed it while `Authorize` returned success, and the caller's first read got a bare `os.ErrClosed` — indistinguishable from reading an upload they closed themselves. The upload now records why it closed and `Read` reports that, so a cancelled transfer surfaces as `context.Canceled`. The window itself cannot be removed (a context can be cancelled at any point after a successful return); what it can do is say what happened. Existing test tightened from "read fails" to "read fails with `context.Canceled`", mutation-checked, `-race` clean.

**`.roborev.toml` guidelines 14 and 15.** The High finding here — "production safeguards are not wired into provider uploads" — restates this PR's declared scope: *"There is no standalone CLI in this PR. Part of #176 (R2). Stacks on #189."* Stacked work lands the capability before the caller by design, so unwired API is the plan, not a gap. Guideline 14 records that. Guideline 15 records that the content scanner is best-effort by design, so each newly imagined attribute or filename spelling is not a separate vulnerability; both keep the "DO flag" half so real regressions still surface.

**Declining the rest.** EPUB manifest duplicates and OOXML content-type part selection are the same reference-enumeration theme, now on its fifth round. PDF `SubmitForm` is on its third appearance and still has no threat model attached. All are better as issues than as more edits here.

Closing out on my side. Round 1 found seven real defects and they are fixed. Since then, three of the four rounds found a regression I introduced in the previous round rather than a defect in the original change, which is the signal to stop editing rather than keep going. Nothing imports `document/upload` outside its own tests, so the residual is not reachable today.

wesm:
Fixed the two Highs and the UNC one in `629148d`. Declining the other three.

All three reproduced before the change, and all three were gaps in my own earlier work rather than new ground:

- **OOXML declared parts** — I fixed selection-by-filename for EPUB in `3f37544` and left OOXML on filenames. A part named `report.bin` declared as `application/xml` in `[Content_Types].xml` was never scanned. Parts are now mapped to their declared content type (Override wins over Default), same shape as the EPUB manifest fix.
- **EPUB rootfiles** — my code read `Items[0]`. It now walks every rootfile a container lists, merging each rendition's manifest declarations.
- **UNC backslash** — `//host/path` was rejected while `\\host\path` was read as relative. One-line fix in `isExternalXMLURI`.

Filename matching stays alongside the declaration in both containers, so an undeclared part is still covered by its name; neither signal alone covers both cases.

Verified: all three closed, ordinary relative references still eligible, and real LibreOffice `.epub`/`.xlsx`/`.ods` still clean — that last check is what caught the `A1:C3` and `opacity="0%"` false positives in earlier rounds. Regression tests added, both mutation-checked, suite and lint green.

**Declining, unchanged from earlier rounds:**

- *Compound-file nesting* — fourth appearance. Matching OLE magic would reject ordinary XLSX/PPTX carrying embedded objects at `xl/embeddings/oleObject1.bin`, which is the same false-positive trap that cost two rounds already. Needs a decision about that tradeoff, not a reflex.
- *`xml:base` in the manifest* — no reachable path shown, and failing closed on it risks rejecting real books.
- *Provider metadata discarded* — a fair observation (accepted, hashed, bytes dropped), but it is a question about the #189 provider contract, not a defect in this branch. Worth its own issue.

wesm:
All three fixed in 9d61117.

- **OOXML counts** — I extended part *selection* by declared type last round but not the *measurement* alongside it, so a declared worksheet outside `xl/worksheets/` was scanned but counted zero sheets and cells. Both now key off the same `isWorksheet`/`isSlide` test. Same for slides.
- **EPUB conflicting declarations** — a later rendition can no longer demote an entry an earlier one put in scope.
- **PDF file specifications** — `/Type` is optional, so a `/F` or `/UF` *string* decides. An integer `/F` (annotation flags) and an `/EF` entry (embedded, not referenced) stay eligible; both have regression tests.

I also audited every remaining filename-keyed site rather than wait for this to surface a fourth time. The only one left is ODF `content.xml`, where the part name is normative in the OpenDocument spec — the filename *is* the declaration there, so it is not the same bug.

Each fix was mutation-checked (revert it, confirm the new test fails), and real LibreOffice `.xlsx`/`.ods`/`.epub`/`.pdf` files stay eligible with unchanged counts.

wesm:
Both confirmed by probe and fixed in ee0c093.

**EPUB conflicting declarations** — right call, and my last fix had the wrong shape. Keeping the first inspectable declaration also mis-scanned in the other direction: a CSS body declared XHTML by some rendition was parsed as markup and rejected as `malformed_input`. Each declaration is how *some* reader reads those bytes, so all of them are now kept and every applicable scanner runs. That deletes the tiebreak rather than patching it. Covered in all four orderings.

An OOXML `Override` still replaces the `Default` for its part — OPC states one content type per part, not a set of alternatives — and that asymmetry now has its own test.

**PDF `/DOS`, `/Mac`, `/Unix`** — real gap. A specification can carry them with no `/F`, and all three were eligible before. Now all five path keys are read, still behind the string-type guard, so an integer `/F` and an `/EF` embedded file stay eligible.

Each fix was mutation-checked, and real LibreOffice `.xlsx`/`.ods`/`.epub`/`.pdf` files are unchanged.

wesm:
Three fixed in 652ecb1, one declined.

**Drive paths** (High) — confirmed. `C:\secret.txt` parsed as a one-letter scheme and passed; `\\host\share` was already rejected, so this closes the same family. A one-letter prefix with no separator after it (`c:notes`) stays local, and that negative control is tested.

**Worksheet/slide counting** — confirmed, and it was a false-positive bug against real files, not just a bypass: Excel writes per-sheet `_rels` parts under `xl/worksheets/`, so a genuine workbook counted 2 sheets per sheet and could fail a limit it actually met. A declaration now settles the type in both directions; only an undeclared part falls back to its name.

**Parameterized media types** — confirmed. `application/xhtml+xml; charset=utf-8` matched nothing, so the resource went unscanned. Normalized with `mime.ParseMediaType` for both EPUB manifests and `[Content_Types].xml`.

**ZIP central directory before `MaxEntries`** — declining, third filing. The archive is fully read under `MaxSourceBytes` before `zip.NewReader` sees it, so the directory is a subset of already-bounded bytes and entry count is capped at roughly `MaxSourceBytes/46`. Allocation is a small constant multiple of a limit the operator already sets; an uploader who can send a `MaxSourceBytes` file can cause that allocation regardless. No capability crosses a boundary that wasn't already crossed. Enforcing `MaxEntries` earlier would mean hand-parsing the EOCD ahead of `archive/zip` — new parsing code in the same trust position, to bound something already bounded.

Each fix mutation-checked in both directions where it has a false-positive edge; real LibreOffice files unchanged.

wesm:
Fixed in 4603ab5.

**xml:base** — confirmed, and broader than reported: `\\host\share\` was accepted too, since the base check had its own `//` test and never got the UNC or drive rules. Root cause is that `xml:base` was classified by an inline check instead of `isExternalXMLURI`, so every rule added to the shared classifier missed it. Now routed through the one classifier, with a relative-base case kept as a control.

Two cleanups from a pass over the accumulated diff:

- `inspectZIP` had reached 160 lines holding byte bounds, part classification, reference scanning, and unit counting in one body — that mixing is what let the counting rule and the selection rule drift apart two rounds ago. Split into `boundZIPEntries` and a `zipContainer` with one function per question. Behaviour-preserving; the existing suite plus real LibreOffice files verify it.
- Removed a local `slicesContains` wrapper that only forwarded to `slices.Contains`.

I did not enable `funlen`: `.golangci.yml:97` disables the complexity linters deliberately ("trust reviewer judgment"), so the split stands on its own merits instead.

wesm:
Real, fixed in 296d999 — with one correction to the finding.

**`../../outside` does not escape.** From `OPS/text/chapter.xhtml` it resolves to `outside`, still inside the container. Whether a reference leaves depends on the entry's depth, so the rule resolves against the entry's own directory instead of matching on `..`. The same `../../outside` in the package document at `OPS/content.opf` does escape, and is now rejected. Matching the text alone would have rejected the ordinary `../Images/cover.png` that real books carry — that case is a test.

**Not adopted: resolving every reference against archive membership.** That is what the CSS scanner does, and it works there because a `url()` is always a locator. Applied to XML attributes it would reject `title="Cover Image"` and every other non-reference value, since the scanner treats all unlisted attributes as potential locators. Containment is checkable without knowing whether a value is a reference; existence is not.

**The manifest half is settled by the same rule.** An href resolving outside the archive was dropped when building the declared-type map, on the grounds that scanning the package document would classify it. That held only for host-bearing values; it now holds for this form too.

**Known limit, unchanged:** an absolute path such as `/etc/passwd` in markup stays eligible. It is ambiguous with an OPC absolute part name (`Target="/xl/styles.xml"`), and URL semantics resolve a leading `/` against the container root rather than the filesystem, so traversal is the form that actually escapes under naive on-disk resolution.

Mutation-checked in three directions: rule removed, depth ignored, backslash normalization removed. The middle one is the false-positive guard. Real LibreOffice files unchanged.

wesm:
Fixed in 2fd3e11: the two Medium findings, the encoded-traversal part of the first High, and one false positive the review missed. Declining the second High with evidence.

**Undetectable formats (Medium)** — real. `.pptm`/`.xlsm`/`.odp` were routed into container inspection with no detector candidate, so a well-formed file of those types always returned `malformed_input`: blaming the document for a format the pipeline does not support. Now reported as an unsupported family. Adding real support is a separate decision, especially for macro-enabled formats.

**xml-stylesheet (Medium)** — real. Every instruction was rejected, including one naming a stylesheet inside the package. The href is now read and classified like any other reference.

**Also found while fixing that:** every `<!DOCTYPE ...>` was rejected as an external reference, including the bare `<!DOCTYPE html>` that nearly every XHTML chapter in a real EPUB carries. That is a larger false positive than anything in this report — most real EPUBs would have been rejected. Only an external identifier or an internal subset is rejected now.

**Archive boundary (High)** — partly real. Percent-encoded traversal (`%2e%2e/`, `..%2f`) did escape, and decoding now happens before containment is measured. The rest of that finding does not hold: drive paths and UNC paths are already rejected (652ecb1, 4603ab5, with tests). A single leading backslash normalizes to a rooted path, which is the absolute-path limit documented on the previous commit, not a new gap.

**Upload fingerprints (High)** — declining; the verification the finding asks for already happens.
- `InspectCapability` seals descriptor, profile, and disclosure fingerprints into the record, and `Checksum` covers them.
- `Authorize` re-inspects the spool and requires `reflect.DeepEqual(reinspected, capability)` (spool.go:261).
- `validateAuthorizationWithoutUpload` compares `DescriptorFingerprint` and `PolicyFingerprint` against the live descriptor (provider.go:906-910), before `provider.Render`.
- The upload/authorization match requires equal capability checksums (provider.go:889).

Copying the fingerprints into `AuthorizedUploadMetadata` would add a second path that can disagree with the sealed record. Exploiting the current one needs a SHA-256 collision.

Five mutation checks, each caught by its own case, including the two false-positive guards. Real LibreOffice files unchanged.

wesm:
Three fixed in be68568, one declined.

**xml:base and containment (High)** — real, and the sharpest finding in a while. Containment was measured from the entry's directory while references resolved from the base, so `xml:base="../"` plus `src="../../outside"` crossed the root and read as local. The two now share one origin that moves with the base. Tested in both directions, including a base that *descends* and therefore requires more distance to leave.

**Encoded OOXML part names (High)** — real. `/xl/re%70ort.bin` matched no entry, so the part was left undeclared and inspected only by its filename: the same bypass reading declarations was meant to close, reachable by respelling the part name. Both spellings now map to the part.

**localAuthority across decoding (Medium)** — real, latent. Decoding into a value that already held authority kept it while replacing every field around it, so a record could describe foreign bytes and still claim upload authority. No caller does this today, but the type documents the guarantee, so it now holds regardless of how it is called.

**Root-relative paths (High)** — declining, and the stated mechanism is incorrect. `path.IsAbs("/etc/passwd")` returns true; the value is not accepted because detection fails, but because `escapesArchive` deliberately returns false for rooted paths. That decision is documented on 296d999: `/xl/styles.xml` is a legitimate OPC absolute part name, and URL semantics resolve a leading `/` against the container root rather than the host filesystem. The suggested alternative — "require an existing member" — is the archive-membership rule I declined on 296d999, because the XML scanner treats every unlisted attribute as a possible locator and `title="Cover Image"` resolves to no member.

There is a narrower option I have not taken: reject rooted references in EPUB only, where OCF requires relative paths, while leaving OOXML alone. I have not done it because last round's `<!DOCTYPE html>` bug showed what over-rejection costs, and I have no corpus of real EPUBs to test it against. Happy to do it if you want that trade.

Four mutation checks, each caught by its own case, two of them false-positive guards. Real files unchanged.

wesm:
Three fixed in fc1984f, one declined. All three real ones were off by exactly one directory level.

**xml:base as a directory (High)** — real. A base names a document, so references resolve from the directory holding it. Treating the base itself as a directory added a level, so `xml:base="a"` bought one step of escape for everything after it. `path.Dir` handles both spellings: `assets/` descends, `assets` does not.

**Relationship origin (High)** — real. A `.rels` part describes another part, and its targets resolve from that part's directory, not from `_rels`. Same extra level. The false-positive risk here is the real one, so it is tested directly: a real package points from `xl/_rels` at `worksheets/sheet1.xml` and from `xl/worksheets/_rels` at `../media/image1.png`, and both stay eligible.

**Absolute manifest href (High)** — real. `path.Join("OPS", "/chapter.dat")` produced `OPS/chapter.dat`, so the declaration named an entry that does not exist while the real root entry stayed undeclared and unscanned. An absolute reference inside a container names the container root, and is now resolved that way rather than rejected, which is what closes the bypass.

**Standalone absolute paths (High)** — declining. This is the rooted-path decision from 296d999 and be68568 in a third form. When `home` is empty there is no container, so containment has nothing to measure; rejecting `/etc/passwd` there would be the absolute-path rule I declined twice, applied to the one case where no container exists to escape. The reasons are unchanged: `/xl/styles.xml` is a legitimate OPC part name, and a leading `/` resolves against the container root under URL semantics.

Five mutation checks. One of them — reducing every base to its parent unconditionally — changed no test, which showed my trailing-slash guard was dead logic rather than showing a test gap; `path.Dir` already normalizes both spellings, so the branch is gone. Real files unchanged.

wesm:
Two fixed in 32a0c9f, one declined for the sixth time.

**Windows rooted path (High)** — real, and this corrects an earlier judgement of mine. I declined `\Users\victim\secret.txt` twice as equivalent to `/etc/passwd`. They are not equivalent: `/etc/passwd` stays eligible because it collides with a legitimate OPC part name like `/xl/styles.xml`, and no such collision exists for the backslash spelling. Both neighbouring spellings, `C:\secret.txt` and `\\host\share`, were already rejected, so allowing the middle one was arbitrary. A further separator is required, so `\frac{1}{2}` in an alt-text attribute stays local — that control is a test, and the over-aggressive mutation fails it.

**Stylesheet pseudo-attributes (Medium)** — real. The scan searched for `href` by name, so a decoy inside an earlier value could point it at a harmless stylesheet while the real external one, later in the same instruction, went unread: `alt=" href='style.css'" href="https://…"` passed. Pseudo-attributes are now read in order as pairs, every `href` is checked rather than only the first, and an instruction that cannot be parsed is unresolvable rather than clean.

**Nested compound files (Medium)** — declining; this is the sixth filing. Matching the OLE signature would reject ordinary spreadsheets and decks, which legitimately carry embedded objects at `xl/embeddings/oleObject1.bin`. Nested ZIP containers are already detected by signature regardless of filename. An embedded OLE payload is opaque bytes that inspection does not dereference and the provider contract does not execute, so no capability crosses a boundary; rejecting it would remove real documents to no effect. Recursive inspection of arbitrary compound files is a different feature, not a fix.

Five mutation checks. Two initially proved nothing — one broke the build, one was equivalent to the real code — and both were redone; the second exposed a missing case, an instruction whose syntax breaks after a valid local href, which is now a test.

wesm:
Both real, fixed in 1b44d95.

**Encoded Windows paths (High)** — real, and the inconsistency was mine. Traversal was already measured after percent-decoding; the drive, rooted-path, and protocol-relative rules were applied only to the raw text. So `C%3A%5Csecret.txt` passed while `C:\secret.txt` was rejected. All of them now go through one check that considers both spellings. `0%` and `100% cover`, whose percent signs are not escape sequences, stay local — the over-aggressive mutation fails on those.

**Manifest xml:base (High)** — real, with a correction to the report's mechanism. My probe showed the resource is unscanned *with or without* `xml:base`, because an entry no manifest declares is never scanned. That part is correct behaviour: no reader loads an undeclared resource either. The actual divergence is narrower — a reader that honours `xml:base` resolves the item to a path my declaration map never held.

Rather than pick a side on whether `xml:base` applies here, the item is now declared at **every** path a reader may resolve it to. Declaring a path that holds no entry is inert; missing the path a reader uses leaves a reachable resource unscanned. The package directory is always among them, so an item with no base resolves exactly as before — that is a test, and dropping it fails four existing cases.

Four mutation checks, two of them over-aggressive variants that the false-positive controls catch. Real files unchanged.

wesm:
Fixed in 1bbb6a1.

**Encoded `xml:base` (High) — real, half of it.** The reference was decoded before containment was measured; the base that moves the origin was not. Dot segments in a base were already handled, so only the encoded spelling (`%2e%2e/`) got through. Both now go through one shared normalizer rather than two copies that can drift — that drift is what caused this and the two before it.

**`alttext` (Medium) — real, but not via the given example.** `\frac{1}{2}` is not rejected; a passing test asserts it is eligible (no separator follows the leading backslash). `\left(\frac{a}{b}\right)` is rejected, and that is a genuine false positive on MathML alt text, which is never dereferenced. `alttext` is now exempt; the neighbouring `altimg` does name an image and stays checked.

**Nested OLE/CFB (Medium) — declining, 7th filing.** Matching CFB signatures inside archive entries would reject ordinary spreadsheets and decks that carry `xl/embeddings/oleObject1.bin`. Nested ZIPs are already caught by signature. Embedded OLE bytes are stored, never dereferenced by the inspector, and never expanded.

wesm:
One of four is real, and it is not a High. Fixed in 1cab6b5.

**Containment misses `..` — real, severity wrong.** `path.Clean` spells the root's own parent as `..`, with no trailing separator, so the `../` prefix test read that one landing place as contained. `../../..` from a two-deep entry reaches it. But that value names a *directory*, not a file — there is no `../../../etc/passwd` form of it, since anything with a name after the dots gets the separator and was already rejected. Nothing is readable through it and nothing dereferences it. Fixed because the predicate answers one question and got one input wrong, not because anything escalated. The three sites now share one predicate.

**PDF `/FS` — not valid PDF.** In ISO 32000-1 Table 43, `/FS` is the *file system* entry of a file specification and is a name object (`/URL`). It never holds a path; the path entries are `/F`, `/UF`, `/DOS`, `/Mac`, `/Unix`, which `detect.go:341-360` already checks as strings. `/FS (/etc/passwd)` describes a file that does not parse as a filespec. The other two suggestions are already done: `pdfDictName` calls `context.Dereference` (`detect.go:383`), and `pdfHasExternalReference` walks every dict in the catalog graph and passes each to the filespec check (`detect.go:275-317`), so nesting is covered.

**CFB v4 — premise inverted.** MS-CFB §2.2: when Major Version is 4 the header is padded with zeros to a full 4096-byte sector, and §2.6 states sector 0 begins at byte offset Header Size, not 512. So a v4 file *is* a whole number of 4096-byte sectors and `(sector+1)*sectorSize` is the correct offset. The proposed `size-512` change would break valid v4 files.

**`localAuthority` forgery — repeat filing, no boundary.** A caller outside `media` cannot set the unexported flag; only `InspectCapability` does, on bytes it just read. Recomputing the checksum over mutated fields does not help, because `Authorize` re-derives from the actual upload: `spool.go:261` compares measurements by value, `provider.go:889` requires the checksum to match bytes the provider hashed itself, and `provider.go:906-910` requires the descriptor and policy fingerprints to match. A forged record authorizes an upload of exactly the bytes it truthfully describes. An in-process caller who can call the inspector on bytes already has that capability before the mutation, so nothing is gained.

wesm:
The High is real and worse than described. Fixed in 7061dae.

**Rooted `xml:base` — confirmed, plus a second symptom the finding missed.** `resolveArchiveDir` treated a base beginning with `/` as naming nothing inside the container and returned no origin. Two failures came from that one answer, both reproduced before the fix:

1. *Manifest mapping* (the reported one). `<package xml:base="/">` with `<item href="chapter.dat" media-type="application/xhtml+xml"/>` left the entry at `chapter.dat` undeclared, so it fell back to its file name, was not read as markup, and an external reference inside it passed as **eligible**.
2. *Content documents* (not reported, and the worse of the two). No origin means containment is not measured at all. A chapter carrying `xml:base="/"` accepted `../../../../secret` — a reference that walks out of the archive — as **eligible**. Without the base, the same reference is rejected.

A rooted base now resolves against the container root and inspection carries on, which is what a reader does. Only a base carrying a scheme still clears the origin. Four regression tests, all confirmed discriminating by reverting the fix.

**`\frac{1}{2}` — declining, third filing of the same example.** It contains no `/` and no second `\`, so `hasRootedBackslash` returns false and it is eligible. A test has asserted that since `32a0c9f`, and I re-ran it against `1cab6b5` before answering. The real MathML carrier, `alttext`, was exempted in `1bbb6a1`. What remains is TeX with two or more macros in an attribute that is not on the vocabulary list, such as `content="\left(\frac{a}{b}\right)"`. I am not tightening the predicate for it: `\Users\victim\secret.txt` and `\left(\frac{a}{b}\right)` have the same shape, so "match a Windows path shape" cannot separate them, and any rule that admits the TeX admits the path.

wesm:
All three are real. Fixed in f733020, with two of the three narrowed from what was reported.

**Base normalization order — confirmed, and the diagnosis is more specific than "wrong order".** The last segment of a base is dropped because it names the document rather than its directory, but `..` is a step through the tree, not a name. From an entry at `OPS/text`, `xml:base="../.."` moved the origin one level instead of two, so `src="../outside"` spent a level that was never taken and read as contained. Same for `xml:base=".."` and `xml:base="a/.."`. The suggested fix — join and clean first, then take the directory — would have broken the trailing-slash case, since `path.Clean` strips the separator that distinguishes `assets/` from `assets`. The fix instead drops the last segment only when there is a name to drop.

**Backslash spellings — confirmed for `/\`, declining `\secret.txt`.** `/\Users\victim\secret.txt` and its encoded form were eligible while the mirrored `\/host/share` was already rejected, which is one half of an idea being caught. A separator-normalizing consumer reads both as `//host`, so both now name a host. `\secret.txt` is a different case and stays eligible: normalized it is `/secret.txt`, a container-root path, which is the same shape as the legitimate `/xl/styles.xml` this PR deliberately accepts. Rejecting it would also reject a single TeX macro, which the previous round asked me not to do.

**ODF manifest — confirmed.** ODF states each part's type in `META-INF/manifest.xml`, the way OOXML uses `[Content_Types].xml` and EPUB uses the package manifest. Only the other two were read, so an ODF part fell back to its file name: an SVG declared under `Pictures/image.dat` was never read as markup and the external reference inside it was never seen. Now parsed, including percent-encoded `full-path` values. Not doing the fail-closed half — an entry nothing declares as markup still resolves by name, since failing closed on unclassified entries would reject ordinary spreadsheets. Real ODS files declare every part here, including `content.xml` as `text/xml`, and stay eligible.

Four mutation checks, each confirmed to fail the intended test.

wesm:
The High is real. Fixed in 512bd1b. Declining the Medium, with a measurement.

**Intermediate base directories — confirmed.** A package document carries `xml:base` in three places: the package element, the manifest, and the item. Only the two ends of that chain were recorded. I reproduced it with `<package xml:base="a/"><manifest xml:base="b/"><item href="chapter.dat" xml:base="c/">`: entries at `OPS/chapter.dat` and `OPS/a/b/c/chapter.dat` were rejected, while `OPS/a/chapter.dat` and `OPS/a/b/chapter.dat` — the two steps in between — were **eligible**, because the item was declared nowhere and fell back to its file name. Every step is now recorded, which is the same choice already made when the two ends were recorded together: inspection does not need to settle which reader is right, only to have scanned whatever any of them loads. Mutation-checked by restoring the two-ends form; both intermediate cases fail.

**ZIP central directory — declining, 4th filing, now with a number.** I built the worst-case archive (the most entries a byte budget can declare: stored, empty, one-character names) and measured `zip.NewReader`:

```
source=64 MiB  entries=713925  heap after NewReader=133 MiB  amplification=2.1x
```

The source bytes are already fully resident before this point — `InspectCapability` reads them under `MaxSourceBytes` at `inspect.go:148-153`, and `validateInspectionPolicy` caps that at 1 GiB. So an attacker who sends N bytes causes about 2.1N more, a constant factor on a budget the operator already chose, not an amplification. There is no bomb here: the central directory cannot declare more entries than 46 bytes each will fit in.

The suggested remedy is worse than the finding. Preflighting the central directory means parsing it, and "a bounded shared ZIP parser" means replacing `archive/zip` with a hand-rolled one — a large new correctness and memory-safety surface to save a 2.1x constant. If 3x peak is too much for a deployment, the lever that already exists is `MaxSourceBytes`.

wesm:
Two of four fixed in 37dd3f8. One declined as a repeat; one declined because the suggested change would create a bypass.

**Standalone absolute paths — real, and my four earlier declines were wrong about this case.** I declined `/etc/passwd` before on the grounds that it collides with a legitimate OPC part name such as `/xl/styles.xml`. That argument holds *inside a container*, where a rooted path names the container root. A standalone XML document has no container, so there is no root for the path to name and it pins a location on the host — exactly what `C:\secret.txt` and `\Users\victim\secret.txt` do, both of which were already rejected. Reproduced: `<doc><image href="/etc/passwd"/></doc>` as `application/xml` was eligible. Now rejected, while a relative reference from the same document stays eligible (it lands wherever the consumer put the document and pins nothing) and a rooted part name inside a container stays eligible. Both directions are tests.

**Declared part names with dot segments — real, in both formats.** `full-path="Pictures/./image.dat"` and `PartName="/xl/./media/i.dat"` named the right part for a reader while matching no entry here, so the part was left undeclared and inspected by its file name. `../` spellings too. Fixed for ODF and OOXML through one shared function — they had already drifted once, when one decoded percent escapes and the other did not.

**`\secret.txt` — declining, 3rd filing.** Answered in full on `f733020`: normalized it is `/secret.txt`, which is the container-root spelling this PR accepts on purpose, and rejecting every lone leading backslash also rejects a single TeX macro, which an earlier round specifically objected to.

**Declarations overriding suffixes — declining, and the suggested change would open a bypass.** The current rule is a union: a part is markup if its suffix says so *or* a declaration says so. Making declarations take precedence lets an attacker suppress inspection of any part by declaring it binary. I checked: `/xl/evil.xml` declared as `image/png` or `application/octet-stream` is still inspected today and still rejected. The failure mode the finding describes runs the other way — over-inspection can cause a false rejection, never a bypass — and I could not construct a real package that hits it, since OPC parts named `.xml` are XML.

Four mutation checks; three discriminated. The fourth showed a guard in the new function that changed no behavior, so I removed the guard rather than write a test asserting nothing.

wesm:
Two fixed in d834f30, two declined.

**HTML `<base href>` — real, and the most serious one found so far.** Only `xml:base` was modelled. `<base>` sets the base URI of the *whole document*, so a renderer resolves every reference against it. Reproduced: a chapter at `OPS/text/` with `<base href="../../">` was **eligible**, because the renderer resolved outside the container while containment was still measured from `OPS/text`. The base is now read in its own pass before scanning — that is what makes a document-wide effect representable in one forward pass, so a reference written *above* the `<base>` element is measured against it too, as a renderer does. The base is a locator itself and resolves from where the document sits rather than from its own answer, so `<base href="../../../../">` is still rejected while `<base href="../../">` stays eligible. Both directions are tests.

**Duplicate `Override` — real.** With `/xl/report.bin` declared `application/xml` then `application/octet-stream`, last-write-wins made it eligible; the reverse order rejected it. So the declaration order decided, and a consumer taking the first reads as markup what we skipped. The part is now inspected under every declaration naming it. An `Override` still replaces the `Default` for its part, which is what separates the two cases, and that has its own test.

**Percent-encoded scheme — declining.** `https%3A%2F%2F169.254.169.254%2Flatest` is not an absolute URI. RFC 3986 §3.1 delimits a scheme with a literal `:`, and percent-decoding happens after parsing, not before; WHATWG URL agrees, since `%` is not a valid scheme character. Both resolve the value as a relative path, so a conforming consumer fetches a path inside the container, not the metadata endpoint. The spellings that a parser *does* read as absolute are already rejected — I checked `https:%2F%2F169.254.169.254/latest` and `//169.254.169.254/latest`, both `external_reference`. This differs from the backslash case I did fix, where WHATWG URL *mandates* the normalization. What would change my mind is a named consumer that decodes before parsing; I could not find one.

**`\secret.txt` — declining, 4th filing.** Same answer as on `f733020` and `37dd3f8`: normalized it is `/secret.txt`, the container-root spelling this PR accepts on purpose, and rejecting every lone leading backslash also rejects a single TeX macro.

Four mutation checks, all four discriminating.

wesm:
All three are real. Fixed in 45ee759. Two of them are follow-on defects in the fix I pushed one commit ago, which is worth saying plainly.

**`<base>` exemption too wide — real, though not by the stated mechanism.** The finding says references *after* a climbing base escape; they do not, and a test has covered that since d834f30. What does escape is markup nested *inside* the element: `<base href="../../"></base>`. An XML parser accepts that nesting even though HTML does not, and the child inherited the unshifted directory. The exemption now covers the base's own href alone, which is all it was meant to cover.

**Namespaced `href` — real.** `<base x:href="assets/">` shifted the origin even though a renderer ignores the attribute. A namespaced base that *descends* is the harmful direction: it left every later reference measured with more room to climb than it has, so `../../../secret` read as contained. Confirmed against the no-base control, which rejects it. Only the unqualified `href` states the base now, and an element named `base` in another vocabulary no longer counts as one either.

**Duplicate `Default` — real, and I missed it.** I fixed exactly this for `Override` in d834f30 and did not check its sibling one loop above. Reproduced: `Default Extension="bin"` declared `application/xml` then `image/png` was eligible; reversed, rejected. Parts are now inspected under each declared type, same as Overrides.

Three mutation checks, all three discriminating. Suite, vet, lint clean; real files unchanged.

Two of the three above are my own regressions from the previous commit, and the `Default`/`Override` pair is the second time in this PR that fixing one member of a pair left the other. That is the failure mode worth watching in the remaining review, more than any individual finding.

wesm:
Real, and broader than the `<base>` pre-scan. Fixed in 3d3a4b7.

**Confirmed.** `<base href=" ../../ ">` with `` was eligible; the same base without the spaces was rejected. The diagnosis of *where* is slightly off, though — the two passes were not disagreeing about trimming. Neither trimmed. `path.Join` reads a leading space as the start of an ordinary segment, so `" ../.."` begins with a directory literally named `" .."` and that absorbs one level of climb. The origin stayed put while a renderer moved it two levels up.

**One spelling the finding did not name.** A URL parser also removes a tab or newline from *anywhere* in a reference, not only the ends. `..\t/../../../etc/passwd` was eligible for the same reason — the tab split a segment. That is a plain reference with no `<base>` involved, so trimming only in the pre-scan would have left it.

Both are fixed in the one function that already reconciles the other spellings a consumer resolves (percent escapes, backslash separators, query and fragment), so reference measurement and host classification strip identically. `xml:base` went through that function already and was not affected either way; I checked before and after.

Controls that must not move, all verified: `<base href="assets/">` still eligible, `cover.png` still eligible, a part named `a b.dat` still matched.

Four mutation checks. Three discriminated. The fourth — over-stripping every space — changed nothing any test could see, because interior spaces do not affect where a reference lands. So I deleted the "space inside a name" case I had just written rather than keep a test that asserts nothing, and reworded the comment that implied the guarantee was load-bearing.

wesm:
Two of four fixed in 3f31afc. Two declined, both with a probe rather than an argument.

**Whitespace before a scheme — real.** Confirmed: `isExternalXMLURI(" file:/etc/passwd", nil, "OPS")` returned `false`. A leading space stops `url.Parse` from seeing a scheme at all, so the value resolved as an ordinary relative name. Whitespace was already discarded before measuring containment and before host classification, and this was the third question asked of the same value that was still reading the raw text. All three now read the same reduced text. The location is slightly narrower than reported — the classifier is shared, so the stylesheet instruction was where it surfaced, not where it lived.

**Manifest href canonicalization — real, for two of the six spellings named.** Confirmed with an entry at `OPS/text/chapter.dat`: `href=" text/chapter.dat "` and `href="text\chapter.dat"` left it undeclared, so it fell back to its file name, was never read as markup, and its external reference passed. Both now resolve. The other spellings listed already worked and still do — I checked `?query`, `#fragment`, `text/./chapter.dat` and `text/c%2Edat` before and after.

**ODF manifest namespace — declining, the premise is inverted.** The concern is that a wrong-namespace manifest suppresses declared-type scanning. It does not: Go's `encoding/xml` matches `file-entry` by local name when the field tag carries no namespace, so the declarations are read either way. I ran an ODS with the manifest namespace changed to `http://example.test/wrong`, declaring a neutral-named SVG containing an external reference. Both the correct and the wrong namespace return `external_reference`. Tightening to require the exact namespace would *create* the suppression this describes.

**CFB v4 — declining, 2nd filing, same spec citation.** MS-CFB §2.2: when Major Version is 4 the header is padded with zeros to a full 4096-byte sector, and §2.6 states sector 0 begins at byte offset Header Size, not 512. A v4 file is therefore a whole number of 4096-byte sectors and `(sector+1)*sectorSize` is correct. The proposed `size-512` change would break valid v4 files.

Three mutation checks, all three discriminating. Suite, vet, lint clean; real files unchanged.

wesm:
Three of five fixed in 71a6cec. Two declined with probes.

**PDF `/EF` — real, for one of the three shapes.** I built all three: `/EF` pointing at a missing object and `/EF` that is not a dictionary are both already rejected as `malformed_input` by the parser, so neither is reachable. `<< /Type /Filespec /F (/etc/passwd) /EF << >> >>` was **eligible** — an empty `/EF` names no stream, and a viewer with nothing to open falls back to the path. `/EF` now suppresses the path check only when it is a dictionary that holds something.

**ODF repeat attribute — real.** `x:number-rows-repeated="1"` placed before `table:number-rows-repeated="20000"` counted 200 cells instead of 4,000,000, under a limit of 1,000,000. The attribute is now matched by namespace and exact case; an unqualified one still counts, since a producer that omits the declaration still means the repeat and reading it as foreign would leave those cells uncounted. Where the attribute is genuinely repeated the largest is counted — Go's decoder returns both, so that path is reachable and has its own test.

**Percent-encoded package root — real.** `PartName="%2Fxl%2Freport.bin"` and `"%2fxl/report.bin"` stayed rooted after decoding, and a rooted name matches no entry, so the part was inspected by its file name. The root separator is now stripped from each spelling after that spelling is decoded.

**`\secret.txt` — declining, 5th filing, now with both halves measured.** Standalone, it is already rejected: `<doc><image href="\secret.txt"/></doc>` as `application/xml` returns `external_reference`, as does `/secret.txt`. Inside a container it is eligible, and so are `/secret.txt` and `/xl/styles.xml`, because there a rooted path names the container root — which is what an OPC part name is. Those are the two cases, and both behave as intended.

**`MaxExpandedBytes` / `MaxEntryBytes` ceilings — declining, the cited limits do not exist.** There is no 500 MB or 100 MB constant in this repository; `grep` finds neither. The only hard ceiling is `maxInspectionSourceBytes` (1 GiB) on `MaxSourceBytes`, which bounds the input before anything is decompressed. The remaining limits are set by the operator calling `InspectCapability`, not by the uploader, so nothing an attacker controls crosses a boundary. If a deployment wants a tighter expansion bound it already has one to set.

Four mutation checks. The fourth initially discriminated nothing, which showed the largest-of-duplicates rule was untested rather than wrong; I added the case that pins it and re-ran.

wesm:
Real — this is a follow-on defect in my previous commit. Fixed in 1a7681c, with one part of the recommendation deliberately not taken.

**Confirmed, three reachable shapes.** I built each as a PDF and ran it. All were eligible with `/F (/etc/passwd)`:

- `/EF << /F 6 0 R >>` where object 6 is a dictionary rather than a stream
- `/EF << /F 42 >>`
- `/EF << /Junk 5 0 R >>` — a real stream, but under a key that is not a path key

Each holds something, which is all my last change checked, and each still leaves a viewer nothing to open. `/EF` now counts only when it resolves to a stream under one of the five path keys.

**Not taking "exempt only the corresponding path key."** That part is wrong, and it rejects ordinary documents. ISO 32000-1 §7.11.4 defines `/EF` as carrying *a subset of* the keys F, UF, DOS, Mac, Unix — fewer entries than the specification has paths is the normal case, not a defect. I implemented per-key exemption first and measured it: `<< /F (data.txt) /UF (data.txt) /EF << /F 5 0 R >> >>` — the shape a real producer writes for an attachment — came back `external_reference`. So did the `/UF`-only mirror of it. The file is embedded in both; there is nothing to fetch. One stream therefore covers the specification, and both shapes are now tests.

Three mutation checks, all three discriminating: any-non-empty-`/EF`, a stream under any key rather than a path key, and an entry that need not be a stream. Real fixtures unchanged; suite, vet, lint clean.

wesm:
High is real. Fixed in 6557d59. Medium declined, 5th filing, with the Zip64 half now measured too.

**CSS whitespace — confirmed, including the crafted-entry precondition.** `cssReferenceInArchive` asked its question of the text as written, while every other reference is reduced first. A leading space hid the `//` that makes a reference protocol-relative, and then became the first character of a directory name rather than nothing. With an entry named `OPS/Styles/ /host.invalid/x` in the archive, `url(" //host.invalid/x")` came back **eligible**; without the space it was rejected. The reference is now stripped before both the host test and the path resolution. The crafted entry is part of the regression fixture, so the test fails without the fix.

**ZIP entry limits — declining.** Answered on `512bd1b` with a measurement; the new element here is Zip64, so I measured that specifically rather than assume the earlier number covered it:

```
zip64 EOCD present: true
source=64 MiB  entries=713925  heap after NewReader=133 MiB  amplification=2.1x
```

The archive in that measurement has 713,925 entries, which is past the 65,535 that forces a Zip64 end-of-central-directory record — I checked for the `PK\x06\x06` signature rather than infer it. So the earlier figure was already a Zip64 archive, and Zip64 does not change the ratio: a central directory entry is at least 46 bytes however it is counted.

The source bytes are resident before this point under `MaxSourceBytes`, capped at 1 GiB by `validateInspectionPolicy`, so an uploader who sends N bytes causes about 2.1N more — a constant factor on a budget the operator chose. Capping the entry count before constructing the reader means parsing the central directory ourselves, which is a larger memory-safety surface than the 2.1x it saves.

One mutation check, discriminating.

wesm:
Valid. `xml:Base="images/"` with `src="../../../secret"` was eligible; `XML:base`, `x:base` and plain `base` were already rejected, so only the local-name half was loose. XML attribute names are case-sensitive, so that base relocates nothing and the climb went unmeasured. Now matched exactly.

Checking it turned up the same defect in the HTML element: `<BASE href="images/">` was honoured too. That one is not the same fix. Element names are case-sensitive to an XML parser and not to an HTML one, so a stored document reaching both kinds of reader is genuinely read both ways — making it exact would let `<BASE href="../../">` be honoured by a browser and ignored here. A non-exactly-spelled element may now only remove room, never add it, which both readers agree on.

Three mutations, each discriminating: loose `xml:base` fails the two `xml:Base`/`xml:BASE` cases; `<BASE>` granting room fails "may not grant room"; `<BASE>` ignored outright fails "still removes room".

wesm:
Declining this one, with measurements.

The described behaviour is real but misattributed. PNG bytes at `OPS/img.xhtml` are rejected `malformed_input` **whether or not the manifest declares them** — identical outcome with the `<item>` removed entirely. The declaration is not what causes the rejection; the suffix is, under the standing rule that an entry a reader may parse as markup has to be well-formed markup. Nothing here is specific to declared media types.

The recommended remedy — declarations exclusively, suffix only as fallback — is a measured bypass of the external-reference check:

```
external ref named .xhtml declared image/png   -> eligible=false external_reference   (today)
external ref named .css   declared image/png   -> eligible=false external_reference   (today)
```

Both go eligible under exclusive declarations. `<item href="c.xhtml" media-type="image/png"/>` is one attribute, chosen by whoever wrote the file, and it would switch off scanning of a content document carrying `http://evil.test/x`. Readers dispatch on the spine and the extension as well as the manifest, so trusting the declaration alone lets the document pick which reader we model. The union is deliberate for that reason: a declaration can only ever add scanning, never remove it.

The residual — a mislabelled resource is rejected rather than skipped — is fail-closed and indistinguishable from the hostile case by construction. Relaxing it to "parse, and accept if parsing fails" is worse: Go XML is stricter than an HTML parser, so anything that fails here and renders there becomes an unscanned content document.

No code change.

PR #191, opened 2026-08-24
Bind provider DNS policy to the actual connection

salmonumbrella:
## What changed

Docbank now applies one transport policy to an exact scheme, host, port, CIDR set, proxy mode, and optional TLS roots or SPKI pins. It resolves once per connection attempt, rejects the entire DNS answer set if any address escapes policy, and dials only the validated IPs while preserving the original Host and TLS identity.

Redirects, destination drift, ambient proxies, and unbounded connect or TLS work are refused. The default local GLM-OCR client now uses the same boundary.

## Why

Checking a hostname and then handing it back to the standard dialer leaves a second DNS lookup and ambient proxy behavior outside the reviewed boundary. Resolution must stay bound to the socket that actually receives provider bytes.

## Usage

Build provider HTTP clients from an explicit egress policy. The same policy must approve every resolved address for each connection attempt; retries can move only across that approved set.

There is no user-facing command change.

Part of #176 (R3). Stacks on #190.

### DIFF
--- a/document/glmocr/client.go
+++ b/document/glmocr/client.go
@@ -13,7 +13,9 @@ import (
 	"io"
 	"mime"
 	"net/http"
+	"net/netip"
 	"net/url"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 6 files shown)


wesm:
rebasing

PR #192, opened 2026-08-24
Add durable, revocable processing consent

salmonumbrella:
## What changed

Docbank now records immutable processing grants and revocations against the current vault incarnation, exact profile and disclosure fingerprints, permitted input classes, and retained artifact classes. Expiry and the revocation fence are checked again before provider access and publication, so a later grant cannot revive work leased under older authority.

Every new or restored vault gets a fresh processing incarnation. Backup JSONL keeps prior consent as history, but restored grants cannot authorize new network activity.

## Why

Restoring a vault must not silently restore permission to send its bytes somewhere. Consent needs to be durable enough to audit and narrow enough to fail closed when the endpoint, deployment, disclosure, inputs, retained outputs, expiry, or revocation state changes.

## Usage

Runtime integrations grant or revoke an exact provider operation through the store boundary, then recheck that authority immediately before egress and publication. No operator-facing command is added here; later application surfaces call this contract.

Part of #176 (R4). Stacks on #191.

### DIFF
--- a/internal/store/consent.go
+++ b/internal/store/consent.go
@@ -1,0 +1,426 @@
+package store
+
+import (
+	"context"
+	"database/sql"
+	"encoding/json/v2"
+	"errors"
+	"fmt"
+	"slices"
+	"strings"
+	"time"
+	"unicode/utf8"
+)
+
+var (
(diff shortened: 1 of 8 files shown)


wesm:
not applicable

PR #193, opened 2026-08-24
Add the standard rendition bridge protocol

salmonumbrella:
## What changed

Docbank now has a fixed provider-neutral bridge for hosted and operator-network rendition services. A bridge receives the exact authorized source, returns `source-evidence/v1` synchronously or through a polled job, and exposes artifacts only through the same-origin job route.

The protocol pins one configured origin and named bearer secret, rejects redirects, ambient cookies, templates, and provider-authored URLs, and enforces finite request, job, response, and artifact limits before downloads begin. Payloads are length-checked and checksummed, submissions are idempotent, cancellations are repeatable, and unknown major versions fail closed.

Exact purges of non-legacy rendition attachments or builds also preserve the same source version's independent legacy search authority.

## Why

Broad provider support should not require provider-specific transport inside Docbank core. One narrow bridge contract lets operators add services without widening destination, credential, URL, or output authority.

## Usage

Register a bridge profile with its origin, credential binding, runtime identity, supported formats, and finite limits. The orchestration layer then submits an authorized upload and either accepts the immediate evidence result or polls the returned job on the same origin.

There is no standalone CLI in this PR.

Part of #176 (R5).


wesm:
rebasing

PR #194, opened 2026-08-24
Add auxiliary MD5 fingerprints without changing content identity

salmonumbrella:
## What changed

Docbank now records an auxiliary MD5 fingerprint for original documents and retained sanitized Markdown while keeping SHA-256 as the only content identity. New writes calculate both digests in one bounded pass, and node, version, API, client, and CLI detail return MD5 when it is available.

Existing vaults converge through a supervised resumable backfill that reads loose or packed bytes through the verified resolver. Missing, corrupt, or conflicting content stays incomplete instead of publishing an unverified checksum. Backup and restore preserve completed fingerprints in both SQLite implementations.

## Why

Some external document workflows still use MD5 for comparison and reconciliation. We need to expose that compatibility fingerprint without weakening deduplication, authorization, audit, or blob identity.

## Usage

Use the existing node or version detail surfaces, including `docbank stat`, to read MD5 when the backfill has completed. SHA-256 remains authoritative everywhere else; MD5 is comparison metadata only.

Part of #176 (F9). Stacks on #193.

### DIFF
--- a/cmd/docbank/cli_test.go
+++ b/cmd/docbank/cli_test.go
@@ -686,13 +686,20 @@ func TestJobsShowsDaemonStatus(t *testing.T) {
 	assert.Contains(t, out, "extract:plain-text")
 	assert.Contains(t, out, "running")
 
-	out, err = runCLI(t, "jobs", "--json")
-	require.NoError(t, err)
 	var got api.JobList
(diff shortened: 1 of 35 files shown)


wesm:
rebasing

wesm:
invalid

PR #195, opened 2026-08-24
Capture original document metadata locally

salmonumbrella:
## What changed

Docbank now preserves metadata embedded in original files as durable local evidence. PDF titles, Office authors, email recipients, calendar dates, photo GPS tags, and media tags can be read through `docbank stat` and content-version detail without changing the original or treating provider claims as authority.

Extraction runs locally against verified original bytes. Results are checksum-sealed and immutable for their extractor version; changing the extractor creates a new generation. Attachment-scoped fields stay separate, sensitive values such as BCC and GPS are marked, and extracted metadata never enters provider requests, Markdown, search, or embeddings.

## Why

Useful document metadata should survive independently of filenames and provider output, but it can contain private or misleading claims. Keeping it local, versioned, bounded, and explicitly non-authoritative gives operators the information without widening disclosure.

## Usage

Run the existing `docbank stat` flow or request content-version detail to inspect available original-file metadata. New and backfilled documents populate the same immutable evidence surface automatically.

Part of #176 (F10). Stacks on #194.


wesm:
rebasing

PR #196, opened 2026-08-24
Add a verified local provider for plain-text renditions

salmonumbrella:
## What changed

Docbank now has a private in-process rendition provider for authorized UTF-8 files. It rereads the supplied bytes, verifies their exact length and SHA-256, and emits one stable provider-neutral evidence unit under an explicit 16 MiB profile cap.

Empty input, invalid UTF-8, NUL bytes, cancellation, oversize input, and byte substitution all fail closed. The provider has no network client, credential binding, hosted consent, retained artifact, or filesystem reopen path.

## Why

Plain text should not need a hosted provider or an ambient filesystem path. It still needs the same exact-byte and evidence authority as every other rendition path so local processing cannot become a privileged shortcut.

## Usage

For library use, construct the provider with `plaintext.New(profile)` and call it through `document.RenditionProvider` with an authorized UTF-8 upload.

This PR does not add daemon, API, or CLI selection. The S1–S3 application-surface work will build the process-local runtime registration around this provider.

Part of #176 (R6). Stacks on #195.

### DIFF
--- a/document/plaintext/provider.go
+++ b/document/plaintext/provider.go
@@ -1,0 +1,229 @@
+// Package plaintext implements the bounded in-process UTF-8 rendition provider.
+package plaintext
+
+import (
+	"bytes"
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io"
+	"slices"
+	"strconv"
+	"time"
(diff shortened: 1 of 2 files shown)


wesm:
rebasing

PR #198, opened 2026-08-24
Add broad-format Docling rendition support

salmonumbrella:
## What changed

Docbank now has a fixed-route Docling provider for every format the selected profile declares and the local media inspector can authorize—not just PDF. PDF results can keep complete natural page evidence, including blank pages; other supported families currently publish bounded Markdown with degraded provenance until their natural-unit mapping can be proved locally.

The provider verifies the sealed upload before egress, uses only same-origin async submit, poll, and result routes, refuses redirects and ambient cookies, and stays on the same task after Docling returns an ID. Structured evidence is accepted only from the supported Docling v1 schema; malformed, partial, drifted, or unlocated output fails closed or degrades explicitly.

## Why

Docling supports a broad file surface, but provider acceptance alone cannot define Docbank authority. We should process every locally inspectable and bounded format while refusing to invent page, slide, sheet, or section provenance that Docbank cannot verify.

## Usage

For library use, construct the client with `docling.New(profile, secrets, httpClient)`. An `operator_network` profile can point at Docling Serve over loopback HTTP without a credential; hosted profiles require HTTPS and may resolve a named API-key binding.

Docbank connects to the service the operator runs; it does not install or launch Docling. Daemon, API, and CLI registration is intentionally deferred to S1–S3.

Part of #176 (R7). Stacks on #196.


wesm:
invalid. merging

PR #199, opened 2026-08-24
[F10–E12 04/25] Add the Datalab Convert rendition provider

salmonumbrella:
## What changed

Docbank now has a fixed-route Datalab Convert adapter for every format a profile declares and the media inspector authorizes. It uploads the exact verified source, polls the known request, and maps Marker Document/Page output into page, slide, sheet, or image evidence. Unverified structure falls back to bounded Markdown with degraded provenance.

Provider callback and result URLs are ignored. Hosted profiles require HTTPS and a named API-key binding; redirects, ambient cookies, proxies, oversized responses, partial results, and version drift fail closed. Once Datalab returns a request ID, Docbank stays on that job instead of risking a second paid submission.

## Why

Datalab can cover more formats than a PDF-only adapter, but the same exact-byte, local-inspection, destination, and provenance rules still apply. Ambiguous or partial conversions must not become silent duplicate charges or incomplete durable evidence.

## Usage

For library use, construct the client with `datalab.New(profile, secrets, httpClient)`. The profile pins its HTTPS origin, named API-key binding, locally inspectable formats, and finite source, result, and artifact limits.

This PR adds the adapter contract only. Daemon, API, and CLI registration is intentionally deferred to S1–S3.

Part of #176 (R8). Stacks on #198.

### DIFF
--- a/document/bridge/client.go
+++ b/document/bridge/client.go
@@ -93,12 +93,9 @@ func New(profile Profile, secrets SecretResolver, httpClient *http.Client) (*Cli
 		profile.MaxResponseBytes < 1 || profile.MaxResponseBytes > maxBridgeResponseBytes {
 		return nil, errors.New("bridge: execution bounds are invalid")
 	}
-	isolatedHTTP := *httpClient
-	isolatedHTTP.CheckRedirect = providerhttp.RefuseRedirects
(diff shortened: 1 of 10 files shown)


wesm:
rebasing

PR #201, opened 2026-08-24
[F10–E12 05/25] Run Mistral OCR through the rendition contract

salmonumbrella:
## What changed

Docbank now runs the existing Mistral OCR path through the same provider-neutral rendition contract as Docling and Datalab. It verifies one sealed upload, resolves the named credential per attempt, keeps authority expiry active through upload and retries, and converts Mistral pages into complete PDF evidence plus bounded Markdown and a sanitized receipt.

Docbank counts PDF pages locally before egress. Oversized documents, missing returned pages, malformed output, filtered or compressed authority it cannot inspect, and unsupported non-PDF formats fail closed.

## Why

Mistral should not bypass the exact source, disclosure, output, and evidence rules just because an OCR client already existed. Provider support also cannot be broader than Docbank’s enforceable local unit authority.

## Usage

For library use, construct the adapter with `mistral.NewRenditionProvider(profile, secrets, httpClient)` and pass it an authorized PDF. PDF is the only advertised format in this PR; non-PDF input remains blocked until its local unit bounds are enforceable.

Daemon, API, and CLI registration is intentionally deferred to S1–S3.

Part of #176 (R9). Stacks on #199.

### DIFF
--- a/document/mistral/client.go
+++ b/document/mistral/client.go
@@ -66,6 +66,7 @@ type Result struct {
 	ReturnedModel  string
 	UnitsProcessed int
 	ProviderBytes  *int64
+	ResponseBytes  int64
 	Metrics        RequestMetrics
 }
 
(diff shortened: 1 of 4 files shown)


wesm:
rebasing

PR #202, opened 2026-08-25
[F10–E12 06/25] Make rendition jobs resumable and publication-fenced

salmonumbrella:
Rendition jobs can now survive daemon restarts and transient failures without repeating an uncertain provider submission. One fenced job may serve multiple authorized waiters, but only when its source, provider execution, output policy, and exact retained artifacts agree. Publication attaches the shared result and flips search heads atomically.

Consent is checked at enqueue, restore, egress, and publication. Provider calls run outside the maintenance gate but inside a narrow revocation fence. Removing the selected waiter reselects authority without releasing an active provider claim; orphaned ambiguous calls stop for operator action. Withheld filenames stay out of provider-visible work, execution identities, and durable snapshots. The immutable profile also derives the complete evidence and normalization policy used by every provider.

Schema version 4 is still unreleased, so rendition tables keep only durable relationships and structural shape in SQLite. Versioned Go identities and shared validators own evolving states, provider values, policies, and operational limits. No released vault format changes.

This PR adds provider-neutral orchestration only. Enqueue, API, CLI, and operator surfaces remain deferred to S1-S3.

Part of #176 (R10). Builds on merged PR #201.


wesm:
rebasing

wesm:
Resolved the confirmed lifecycle blockers and the latent worker issues in `f98e38b`.

What changed:

- Re-fence active build/generation roots whenever waiter deletion requeues a shared job, and keep selection on an authorized remaining waiter.
- Keep staged roots alive across reopenable consent/stale-authority failures while waiting consumers remain; reset orphaned staged state and release roots when no waiting consumer remains.
- Move provider execution outside the daemon-wide operation gate while keeping each catalog/blob mutation gated.
- Treat seal-time authorization races and unclassified failures with a durable resume handle as safe transient retries.
- Persist provider submission attempts and apply shared exponential retry delay (1s to a 10m cap) for definitive retryable submissions without `Retry-After`.
- Replace the registry readiness snapshot with provider-neutral late-registration admission, so the worker wakes when the first adapter is registered.
- Centralize staged root refresh and retry policy so future rendition providers reuse the same lifecycle machinery.

Regression coverage now exercises backup/export after waiter requeue, GC after reopenable failure, rejected-sibling root release, provider calls outside the maintenance gate, seal expiry, raw resume failures, retry escalation, durable attempt counts, and late adapter registration.

Verified locally:

- focused store, worker, and daemon regressions
- `go test -count=1 -tags fts5 ./...`
- `CGO_ENABLED=0 go test -count=1 -tags fts5 ./...`
- `make lint`
- `make docs-build`
- `prek run`

I did not change the lower-priority lexical generation-ID or backfill pacing suggestions in this fix, and I did not poll CI.

wesm:
Resolved the follow-up correctness and privacy review in the latest fix commit:

- Publish-time stale-generation, consent, and authority errors now pass through the job-state classifier even when catalog retry handling wraps them. They no longer stop the worker. The stale-generation regression test now exercises the integrated `RunOne` path.
- A transient provider failure without a durable handle returns the job to queued phase and clears its selected waiter and sealed authority. A later claim can select fresh consent. Durable-handle retries continue to preserve the exact resume snapshot.
- Enqueue now rejects captured-artifact cardinalities that contradict deterministic local output or require provider artifacts that the immutable profile does not retain. This fails before provider work begins.
- Provider-specific resume setup receives filename-filtered work and snapshot copies when disclosure is disabled. The worker retains the full identity only for local validation, deduplication, and the sealed resume operation.

I left the two lower-priority tradeoffs unchanged: a fully backed-off auxiliary-checksum page can delay later pages, and reclaim still derives a new lexical generation from the claim epoch. Both are now explicit in the PR description.

Verified with focused regression tests, the full `go test -tags fts5 ./...` suite in CGO and pure-Go modes, the processing race suite, `make lint`, `make docs-build`, and `prek run`.

wesm:
Resolved both follow-up findings in the latest fix commit:

- If the selected waiter's consent changes at the final pre-egress fence, the store now rejects that waiter and atomically requeues the job under another authorized waiter. The worker does not call the provider until the surviving authority is claimed and checked again. If no authorized waiter remains, the existing consent or stale-authority failure is preserved.
- The provider-specific `Prepare` and resume hooks now receive the same disclosure-filtered work copy. When filename disclosure is disabled, neither hook receives the source filename; the complete value remains only in the durable local identity used for validation and deduplication.

The regression coverage exercises both paths through `RunOne`: one provider call publishes for the surviving waiter, while the revoked waiter is excluded; initial preparation and durable resume both observe an empty filename.

I also replaced the PR description with a short, outcome-first version that keeps only the operating guarantees, integration boundary, and known scope tradeoffs.

wesm:
Fixed in `64d3784`.

- Job IDs and database uniqueness now use the same immutable build identity. A second execution identity is rejected at enqueue instead of running until staging collides.
- Jobs persist a narrow artifact-selection fingerprint for provider Markdown, sanitized Markdown, and typed artifacts. Waiters with incompatible retention output cannot share provider work; unrelated profile differences can still share a build.
- A per-vault consent fence now spans the final durable authority check through adapter invocation. Revocation that wins the ordering race prevents provider egress, without holding the daemon maintenance gate during the remote call.

The integrated race test pauses after authorization commits, starts a concurrent revocation, and proves it cannot cross the pre-provider window. Focused suites and race tests, both full SQLite test modes, lint, strict docs, and `prek run` all pass.

wesm:
@salmonumbrella it's nothing you did, but watch out for these clankers adding a bunch of hokey SQLite constraints (especially enum value checking). I'm unwinding some of that now

wesm:
Applied the SQLite-policy cleanup in `3216dde`.

- Removed the PR-local enum checks, late JSON/text size checks, and policy-shaped composite uniqueness constraints from the rendition job tables. Deterministic versioned IDs remain the deduplication boundary, and shared Go validators reject unknown durable values during normal reads, restore, export, and audit.
- Made the captured artifact policy the canonical output-sharing contract. Optional captured roles must match the profile's requested retention, and retained outputs must appear in the policy. Waiters with incompatible output needs therefore get distinct jobs before provider work begins.
- Kept useful limits at actual Go boundaries: captured artifact policy remains capped before persistence, and provider resume handles retain their format and size validation. Execution identity and snapshots stay typed and canonical without a second late SQL byte cap.
- Added repository guidance to treat SQLite constraints as storage compatibility commitments and to prohibit schema enums.

I did not rewrite constraints in already-merged schema surfaces; that would require a separate storage-compatibility decision. The race-enabled store/worker tests, full CGO and pure-Go suites, lint, strict docs build, and `prek` all pass.

wesm:
Cleaned the full unreleased schema-v4 rendition surface in `ce630ad`.

Version 0.14.0 uses schema v3, and schema v4 has not shipped. This is the last clean point to avoid turning provider policy changes into storage migrations:

- Removed SQLite enums and product-policy size caps from rendition profiles, builds, artifacts, units, segments, roots, and loose-retirement encoding.
- Kept primary and foreign keys, immutable identity, ordering, boolean, nonnegative, relational, digest-shape, and fencing constraints.
- Removed the constant physical artifact-state column and its redundant composite unique key. Logical backup metadata still records and validates verified artifact state.
- Kept evolving value and size validation in shared Go validators, including validation when stored profiles and builds are loaded.
- Removed the exact-layout fixture for an earlier unshipped schema-v4 draft. No migration or compatibility path is needed for a layout that never released.

Verified with focused store/processing/backup tests, race tests, the full `fts5` suite under both mattn and modernc SQLite, lint, strict docs, and `prek run`.

wesm:
Resolved both findings in `6b94f04`.

- Confirmed the consent race by strengthening the integrated worker test: on the old code, revocation completed while the provider was blocked inside `Render`. The revocation fence now remains held until `Render` or `RenderResumable` returns. This serializes consent revocation only; the daemon maintenance gate remains released during provider work.
- Added the rendition job tables and indexes to the exact processing-schema identity. The preflight now derives related index and trigger coverage from its expected table set instead of maintaining three duplicate table lists. A malformed `rendition_jobs` definition now fails on open.

Verified the ordinary and resumable provider paths, the malformed-job-schema regression, store and processing race tests, the full `fts5` suite with both SQLite drivers, lint, strict docs, and `prek run`.

wesm:
Resolved all four findings in `2726ab9` after reproducing each failure against the previous head.

- Restore and publication now require every waiter's retained artifact classes to exactly match the shared job's captured policy, so consent cannot under-declare published artifacts.
- Withheld filenames are removed from canonical execution identities and durable snapshots, so provider-invisible metadata no longer splits equivalent work.
- Deleting the selected waiter during an active provider call preserves the claim while authority is reselected. A checkpointed handle cannot be resumed concurrently.
- Complete evidence and normalization policies are derived from the immutable processing profile through one shared helper, then checked at enqueue, restore, and execution.

Verification passed in both supported SQLite modes: the focused regression tests, race-enabled document/processing/store suites, `make test`, the pure-Go full suite, `make lint`, `make docs-build`, and `prek run`.

wesm:
Reviewed both findings.

The durable-handle issue was valid and is fixed in `09d19ad`. I reproduced the full retry → sole-waiter deletion path first: the previous code deleted the job, which allowed identical work to be created again. A provider-phase retry now preserves `provider_started` whenever it retains a resume handle. Deleting the final waiter therefore leaves an operator-required tombstone with the handle intact, and fresh consent cannot resubmit the operation. The new regression exercises that complete path.

I did not add the suggested configuration fallback. `v0.14.0` contains neither processing-profile configuration nor `document/profile.go`; this surface was introduced after that release and remains unreleased. There is no shipped profile shape to migrate, and this repository deliberately does not retain compatibility paths for unshipped contracts. `max_document_chars` remains explicit so the immutable profile fully defines the evidence and normalization identity.

Verification passed: the focused retry/deletion/restore tests, race-enabled store suite, `make test`, the pure-Go SQLite full suite, `make lint`, `make docs-build`, and `prek run`.

wesm:
Confirmed and fixed in `ae3e9e0`.

I reproduced the zero-waiter operator-required path before changing the query: the exact derivative purge scan returned the job's 20-byte source blob as deletable even though ordinary and paged garbage collection retained it. `UnreachableDerivativePurgeBlobs` now applies the same `rendition_jobs.source_sha256` authority check as every other garbage-collection path.

The existing ambiguous-job source regression now covers ordinary, paged loose/packed, and exact derivative garbage collection. Verification passed with the focused purge/resume tests, race-enabled store suite, `make test`, the pure-Go SQLite full suite, `make lint`, `make docs-build`, and `prek run`.

wesm:
Reviewed both findings.

The execution/profile finding was valid and is fixed in `035757d`. Before changing the validator, I reproduced enqueue acceptance for five out-of-profile cases: an oversized source, an oversized response, a substituted artifact role, an omitted requested role, and unrequested provider Markdown. `ValidateRenditionExecutionProfileV1` now binds source and response ceilings plus the exact requested-artifact policy. That shared boundary is enforced at enqueue, metadata restore, resume, and the sealed pre-egress worker check.

I did not add a profile compatibility fallback. `v0.14.0` is still the latest release and contains neither processing-profile configuration nor `document/profile.go`; this canonical profile surface was introduced after that release and remains unreleased. There is no shipped profile format to migrate, and the repository deliberately does not retain compatibility paths for unshipped contracts.

Verification passed: focused document/store/processing/config/backup tests, race-enabled document/processing/store suites, `make test`, the pure-Go SQLite full suite, `make lint`, `make docs-build`, and `prek run`.

wesm:
Confirmed and fixed in `fa038d5`.

The worker now keeps cleanup armed while it owns the prepared upload, including validation, sealing, and authority failures. Immediately before `RenderRenditionWithResume` takes ownership, the worker disarms that cleanup, so the renderer remains the only closer after transfer.

The regression test counts closes across a fresh submission and a definitive retry. Before the fix it observed two closes after the first attempt and four after the second; it now observes exactly one close per upload.

Verification passed: the focused regression, the full processing package, its race-enabled suite, both complete SQLite test suites, `make lint`, `make docs-build`, and `prek run`.

wesm:
Confirmed and fixed in `eb4b268`.

I reproduced the pristine-target gap with stale on-disk operational rows. Before the fix, backup restore accepted both rendition staging and derivative blob-purge authority when the imported stream supplied the matching blob. A stale pack-purge row was rejected only later by the foreign-key audit, after the pristine preflight had passed.

The pristine check now includes all three operational tables and rejects them before deleting the bootstrap root or importing metadata. The regression covers rendition staging, derivative blob purge, and derivative pack purge, and verifies that rejection leaves the original target state intact for inspection.

Verification passed: the focused regression, the full store package, the race-enabled store suite, both complete SQLite test suites, `make lint`, `make docs-build`, and `prek run`.

wesm:
Fixed narrowly in `3cd474f`.

The restore boundary now rejects three inconsistent states before installing the restored database:

- a shared job's captured artifact policy is incompatible with any waiter's immutable profile;
- provider-phase state would permit a fresh submission despite having crossed the prior provider boundary; or
- staged build/generation roots are missing, stale, or point at another job's immutable work.

I added behavioral regressions for all three cases. Each malformed snapshot restored successfully before this change and is now rejected. Existing staged-job round trips still restore and keep their work live.

This is intentionally limited to central restore validation and tests. It does not change the schema, migrations, or the runtime state machine.

Verified with the focused regressions, the full CGO/mattn and pure-Go/modernc test suites, the race-enabled store suite, lint, strict docs build, and `prek run`.

wesm:
Fixed narrowly in `0d92edc`.

I confirmed the failure through the integrated resume path: a durable snapshot whose sealed policy fingerprint differs from the registered provider descriptor is rejected before the provider call. Previously that untyped local error was treated as transient because a durable handle existed, so the job returned to `retry_wait` indefinitely.

The policy mismatch now has a shared document-level error identity, and the worker classifies only that exact error as `operator_required`. The durable handle is retained, no second provider call occurs, and other unclassified resume failures keep their existing transient behavior.

No profile, schema, migration, or general retry-policy changes were made.

Verified with the focused integrated regression, full CGO/mattn and pure-Go/modernc suites, race-enabled document and processing tests, lint, strict docs build, and `prek run`.

wesm:
Fixed pragmatically and narrowly in `8305f95`.

Both findings were reproducible:

- a restored waiter and matching grant using `derived_upload` passed restore even though rendition execution sends the original source; and
- provider-started work relabeled to `queued` bypassed reclaim quarantine and received a new claim without a resume handle.

The fix centralizes the rendition waiter input rule and applies it at enqueue, metadata restore, and the final waiter-authorization lookup. Rendition jobs now accept only the exact `original_file` consent class.

Restore also validates the legal state/phase matrix and the queued/provider boundary. Reclaim quarantines provider-started work without a handle in any phase that could cross provider egress again.

One deliberate exception is retained: `build_staged` and `generation_staged` work remains reclaimable without a provider handle because immutable provider output already exists and recovery performs only local staging/publication. The existing no-second-provider-call regression verifies that boundary.

No schema, migration, or new state was added.

Verified with the focused regressions, full CGO/mattn and pure-Go/modernc suites, race-enabled store tests, lint, strict docs build, and `prek run`.

wesm:
Addressed both findings narrowly in c6fa512:

- Successful publication now retires the selected waiter and consent authority. Waiter-deletion reconciliation does the same for `completed` and `operator_required` jobs whose selected waiter is gone, without changing their terminal state or phase.
- The shared execution-identity validator now requires a filename when disclosure is enabled. Invalid durable snapshots are quarantined as `operator_required` before any resume attempt.

Focused regression tests cover publication cleanup, deletion followed by metadata export, the disclosed-filename implication, and malformed durable-snapshot quarantine. The full CGO and pure-Go test suites, lint, strict docs build, and pre-commit checks pass.

wesm:
Addressed the remaining failed-job reconciliation finding and rebuilt the branch as one commit on current `main` (`0411933`).

- Waiter deletion now includes failed jobs. Consent and stale-authority failures requeue only from a safe staged/queued boundary or with a complete durable resume; terminal and otherwise unsafe failures keep their outcome and provider evidence while releasing the deleted authority.
- Regression coverage proves a failed staged shared job moves to its remaining waiter and a terminal failed job remains terminal while metadata export succeeds.
- The rebase conflict combined `main`'s visual-preview blob authority with this PR's rendition-job source authority, retaining both garbage-collection and restore-verification protections.

The full CGO and pure-Go suites, lint, strict docs build, and pre-commit checks pass. GitHub now reports one commit, the expected `main` base, and a mergeable PR.

wesm:
Applied narrowly and kept the PR as one commit.

- A seal-time provider policy mismatch now moves the job to `operator_required` before egress. Other seal failures, including authorization expiry, remain retryable. An integrated worker test covers both classifications.
- The linked Windows job failed only because the lifecycle-close test allowed one second after the walker released its lease. The assertion remains intact, but now uses the repository's existing five-second lifecycle-close window. The focused test passed 20 consecutive local runs.
- I did not add a `max_document_chars` compatibility default or migration. `processing-profile/v1` was not present in v0.14.0 (`document/profile.go` did not exist in that release), so there is no released profile contract to migrate. Adding a fallback would preserve an unreleased intermediate shape.

Passed: `make test`, `CGO_ENABLED=0 go test -tags fts5 ./...`, `make lint`, `make docs-build`, and `prek run`.

Force-pushed `0411933` → `068c2d9` with the original feature author preserved. GitHub reports the PR open, mergeable, targeting `main`, and containing exactly one commit. A new CI run will provide the Windows-runner confirmation; I have not polled it.

wesm:
Addressed the confirmed recovery issue in `c8b38f2`.

- Fresh consent for a durable provider handle now preserves `provider_started` while replacing only the stale consent authority. The restored job remains valid metadata and resumes the known provider operation instead of looking like a fresh submission.
- The existing end-to-end restore/re-consent test now validates metadata before reclaiming and resuming the handle. It failed on the inconsistent provider boundary before the fix.

I did not add a `max_document_chars` fallback. `processing-profile/v1` did not exist in v0.14.0; the field and its required validation are part of the same unreleased contract. A default would preserve an unreleased intermediate format, contrary to the repository's roll-forward compatibility rule.

Passed the focused regression, the complete store suite, both full SQLite test modes, lint, strict docs, and `prek run`. The PR was fast-forwarded without squashing or force-pushing.

wesm:
Fixed in `369709e`.

The policy mismatch occurs before `BeginRenditionProvider`, so no provider operation is ambiguous. It now records the accepted pre-egress combination `failed/queued` with failure code `terminal`; `operator_required` remains reserved for ambiguous provider outcomes.

The integrated worker regression now exercises the mismatch, verifies that no provider call occurs, exports the resulting metadata, restores it into a pristine store, and confirms the terminal state survives the round trip. Before the fix, the test reproduced the `operator_required/queued` validation failure.

Passed the focused regression, complete processing package, both full SQLite test modes, lint, strict docs, and `prek run`. Pushed as an incremental commit without rewriting history.

wesm:
Investigated; no code change is needed for this finding.

`RenditionJobWorkByClaim` validates waiter structure but does not evaluate live consent expiry or revocation. That live check occurs in `BeginRenditionProvider`. Its existing transaction rejects the stale selected waiter, selects another authorized waiting waiter, requeues under a new claim epoch, and leaves the execution snapshot and durable resume handle intact. Expiry and revocation use the same branch.

I exercised the reported sequence against the real store with two waiters and a checkpointed provider handle: revoke the selected waiter, reclaim the retry, load work, cross the provider boundary, and reclaim again. Work loading succeeded; the boundary returned `ErrRenditionJobWaiterReselected`; the next claim retained the same handle and selected the surviving waiter; its authority check succeeded.

The investigation test passed against `369709e` without production changes and was removed afterward. The worktree remains clean, so there is no new commit or push.

PR #203, opened 2026-08-25
[F10–E12 07/25] Add the self-hosted Marker rendition provider

salmonumbrella:
## What changed

Self-hosted Marker is now a first-class broad-format rendition provider. Docbank uploads the exact authorized bytes to the operator-pinned `/marker/upload` route for PDF, safe still images, DOCX, XLSX, PPTX, EPUB, and HTML/HTM.

PDF and still-image results keep complete natural evidence only when Docbank can prove the local source units match. Transformed document families are marked degraded instead of inventing provenance. Redirects, provider-authored destinations, malformed or partial output, animated images, expired authority, and runtime drift fail closed.

## Why

Marker is useful beyond PDF, but self-hosting should not create a loose local-network trust boundary. The adapter still needs pinned deployment and runtime identity, exact bytes, fixed routes, finite limits, and locally provable evidence.

## Usage

For library use, run the audited Marker service and construct the adapter with `marker.New(profile, secrets, transport)`. The injected transport owns the operator-network destination policy, and Docbank never gives Marker a vault path.

This PR does not add daemon, API, or CLI registration. That application wiring is intentionally deferred to S1–S3.

Part of #176 (R11). Stacks on #202.

### DIFF
--- a/document/marker/client.go
+++ b/document/marker/client.go
@@ -218,6 +218,9 @@ func (client *Client) Render(ctx context.Context, upload document.AuthorizedUplo
 	if metadata.ByteLength > client.profile.MaxDocumentBytes {
 		return document.RenditionResult{}, providerError(document.RenditionErrorPolicyRejected, "Marker input exceeds the document byte limit", nil)
 	}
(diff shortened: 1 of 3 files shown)


wesm:
rebasing

PR #204, opened 2026-08-25
[F10–E12 08/25] Add the verified local PyMuPDF PDF provider

salmonumbrella:
## What changed

Docbank now has a verified local PyMuPDF PDF-text provider without giving an ambient Python interpreter or filesystem surface to the rendition path. It sends the exact authorized PDF bytes over stdin to one operator-pinned executable and accepts only the fixed versioned stdout contract.

Docbank proves the PDF page count locally before publishing complete page evidence. Missing pages, partial output, runtime drift, oversized output, cancellation, expired authority, and executable replacement fail closed. Child stderr never crosses the public error boundary.

## Why

Local PDF extraction should avoid network disclosure without becoming a privileged path that can reopen arbitrary files, discover runtimes, or publish page claims Docbank cannot verify.

## Usage

For library use, construct the provider with `pymupdf.New(profile)`. The profile pins the exact bridge executable, its SHA-256, immutable runtime identity, timeout, and byte/page limits; Docbank does not install Python or pass a source pathname.

This PR does not add daemon, API, or CLI selection. That application wiring is intentionally deferred to S1–S3.

Part of #176 (R12). Stacks on #203.


wesm:
rebasing

PR #205, opened 2026-08-25
[F10–E12 09/25] Add the hosted LlamaParse rendition provider

salmonumbrella:
## What changed

Docbank now has a hosted LlamaParse adapter that receives only sealed authorized document bytes. Structured pages keep page-level evidence; Markdown fallback is marked degraded, and contradictory or partial results fail closed.

Accepted jobs get a durable secret-free resume handle, so a restart can continue the known parse instead of repeating an ambiguous paid submission. Fixed routes, model identity, request and result limits, returned image bytes, redirects, provider-authored URLs, expiry, cancellation, rate limits, and capacity failures all stay inside the shared rendition boundary.

## Why

Hosted parsing needs durable resume semantics and explicit provenance. A transient failure after submission must not cause duplicate work, and provider output must not gain destination or evidence authority just because the service returned it.

## Usage

For library use, construct the client with `llamaparse.NewProvider(profile, secrets, transport)`. The profile pins the hosted identity and finite limits; the injected transport owns the approved destination policy.

Durable resume is consumed by the R10 worker contract. Daemon, API, and CLI registration is intentionally deferred to S1–S3.

Part of #176 (R13). Stacks on #204.

### DIFF
--- a/document/internal/providerutil/multipart.go
+++ b/document/internal/providerutil/multipart.go
@@ -7,7 +7,17 @@ import (
 	"io"
 	"mime/multipart"
 	"net/textproto"
+	"strings"
 )
+
+// ValidateMultipartFilename rejects line breaks before a filename is placed in
+// a multipart Content-Disposition header.
+func ValidateMultipartFilename(filename string) error {
+	if strings.ContainsAny(filename, "\r\n") {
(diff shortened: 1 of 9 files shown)


wesm:
rebasing

PR #206, opened 2026-08-25
[F10–E12 10/25] Add the hosted Reducto rendition provider

salmonumbrella:
## What changed

Docbank now has a hosted Reducto adapter for authorized PDFs and slide decks. Complete output keeps page or slide evidence; URL results, returned PDFs or images, partial output, unsupported spreadsheet provenance, malformed responses, and changed runtime identity fail closed.

Accepted jobs get a durable secret-free resume handle before polling. A restart resumes the known job with cumulative usage history instead of paying for an ambiguous resubmission. The final receipt keeps the original authorized start and accepted-job checkpoint time, so a historical resume stays within its sealed authorization interval.

## Why

Reducto’s async API needs the same exact-byte, destination, evidence, cost, and lifecycle boundaries as every other hosted provider. In particular, an uncertain response must never turn into a duplicate paid parse.

## Usage

For library use, construct the client with `reducto.NewProvider(profile, secrets, transport)`. The profile pins the hosted identity, PDF/PPTX support, and finite limits; the injected transport owns the approved destination policy.

Durable resume is consumed by the R10 worker contract. Daemon, API, and CLI registration is intentionally deferred to S1–S3.

Part of #176 (R14). Stacks on #205.

### DIFF
--- a/document/reducto/client.go
+++ b/document/reducto/client.go
@@ -554,11 +554,11 @@ func naturalEvidence(
 			if err := validateBlockLocation(block); err != nil {
 				return document.SourceEvidenceV1{}, nil, err
 			}
+			if !utf8.ValidString(block.Content) {
+				return document.SourceEvidenceV1{}, nil, errors.New("block content is invalid")
+			}
 			if chunk.Content == "" && block.Content != "" {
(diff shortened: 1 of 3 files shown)


wesm:
rebasing

PR #207, opened 2026-08-25
[F10–E12 11/25] Add the broad-format Unstructured bridge profile

salmonumbrella:
## What changed

Docbank now has a hardened Unstructured bridge profile for the formats it can currently inspect and bound locally: PDF, PPTX, XLSX, ODS, CSV, EPUB, EML, XML, Markdown, plain text, JPEG, and PNG.

The profile pins deployment and runtime identity, exact-byte disclosure, finite source, evidence, Markdown, result, and artifact limits, and the one allowed structured artifact. The generic bridge enforces those ceilings before upload and across inline or separately fetched results. Provider acceptance alone never expands the advertised format set.

## Why

Unstructured supports a broad surface, but “the provider accepts it” is not enough to make the input or output safe durable authority. Docbank should expose the useful intersection it can inspect and bound today, then expand deliberately as local authority improves.

## Usage

For library or adapter deployment, build the canonical compatibility profile with `unstructured.NewProfile(config)`. The operator’s bridge implementation must match that identity and run behind `docbank-rendition/v1`; unsupported upstream formats do not enter the profile automatically.

This PR adds the reference profile, not daemon, API, or CLI registration. That application wiring is intentionally deferred to S1–S3.

Part of #176 (R15). Stacks on #206.

### DIFF
--- a/document/bridge/client.go
+++ b/document/bridge/client.go
@@ -31,6 +31,7 @@ const (
 	maxBridgeTimeout        = 24 * time.Hour
 	maxBridgePollAttempts   = 10_000
 	maxBridgeResponseBytes  = int64(512 << 20)
+	maxBridgeDocumentBytes  = int64(1 << 40)
 	maxBridgeErrorMessage   = 1024
 	maxBridgeCleanupTimeout = 5 * time.Second
 )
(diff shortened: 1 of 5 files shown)


wesm:
rebasing

PR #209, opened 2026-08-25
[F10–E12 13/25] Add the isolated Trafilatura HTML provider

salmonumbrella:
## What changed

Docbank now has a local Trafilatura provider for exact supplied HTML and XHTML bytes. It exposes no URL, redirect, linked-resource, feed, sitemap, remote-metadata, or arbitrary option surface. Accepted output must cover the locally parsed visible text exactly; section provenance is complete only when Docbank verifies the supplied DOM paths and headings itself.

On Linux, the default runner seals the digest-verified bridge executable, launches it inside private user, network, PID, and mount namespaces, and applies no-new-privileges plus a seccomp boundary. IP and Unix sockets, io_uring networking paths, and the amd64 x32 syscall ABI are denied. Output is bounded while streaming, cancellation and overflow tear down the full process tree, and launcher setup failures remain distinct from bridge exit codes.

## Why

HTML extraction should stay local and never turn into ambient web fetching or invented structure. Running a parser binary also needs a real isolation boundary; “local process” is not enough if it can reach the network, host sockets, unrelated processes, or arbitrary files.

## Usage

For library use on Linux, construct the provider with `trafilatura.New(profile)`. The profile pins the exact bridge executable, its SHA-256, immutable runtime identity, and finite limits; the built-in runner supplies the isolation boundary. macOS and Windows require a separately audited injected runner and otherwise fail closed.

This PR does not add daemon, API, or CLI selection. The approved S1–S3 application-surface work will construct and register providers with the R10 runtime registry.

Part of #176 (R17). Stacks on #208.

### DIFF
--- a/document/internal/providerutil/bounded_buffer.go
+++ b/document/internal/providerutil/bounded_buffer.go
@@ -1,0 +1,65 @@
+package providerutil
+
+import (
+	"slices"
+	"sync"
+)
+
+// BoundedBuffer retains at most limit bytes and calls onOverflow once when a
+// writer supplies more. It remains safe while command cancellation and output
+// collection run concurrently.
+type BoundedBuffer struct {
+	mu         sync.Mutex
+	data       []byte
+	limit      int64
+	exceeded   bool
+	overflow   sync.Once
+	onOverflow func()
+}
+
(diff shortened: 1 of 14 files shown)


wesm:
rebasing

PR #210, opened 2026-08-25
Define portable embedding contracts

salmonumbrella:
Docbank now has a provider-neutral contract for generating and storing embeddings. It records the exact model-input convention and vector-space identity so results from incompatible models or settings cannot be mixed.

Providers receive bounded, verified file uploads or normalized text inputs. Their results are validated and copied before they leave the contract boundary. A canonical, checksummed vector format gives stored results one portable representation. Profiles contain credential references only, never secret values.

Processing profiles have not appeared in a Docbank release, so this updates their unreleased representation directly instead of adding a migration for a format users cannot yet have. This establishes the execution and storage contracts; it does not yet schedule embedding jobs or add vector search.


wesm:
rebasing

wesm:
rejected

PR #211, opened 2026-08-25
[F10–E12 15/25] Make embedding inputs tokenizer-specific and frontmatter-free

salmonumbrella:
Embedding inputs are now built from normalized evidence with the binding's own tokenizer, so chunks fit the model's real token limit and YAML frontmatter never reaches embedding text. Before this, the only chunker in the embedding path counted runes and read the legacy normalized document.

## Outcomes

- **One chunk declaration.** The profile's `EmbeddingChunkPolicyV1` is the source of truth. It gains `tokenizer_revision`, and `truncation_policy` is a closed vocabulary. Bindings gain `max_input_tokens` so the chunk budget no longer doubles as the provider limit. `NewInputPolicy` resolves a binding for runtime use and rejects a tokenizer whose identity differs from the declaration.
- **Limits are not identity.** Aggregate ceilings and work budgets live in `GenerationLimits` and stay out of fingerprints. Raising one never invalidates existing generations.
- **Canonical artifact.** Generations encode, decode, and fingerprint through `internal/canonical` like every other durable contract. Forged totals, policy fingerprints, and checksums are rejected on decode.
- **Reviewed profiles** for Nomic, E5, BGE-M3, GTE, Qwen3, and query-only instructions. Model names never select a prefix.
- **Removed** the rune-based plan, recipe, and distillation flow in `document/embedding`. It had no consumer.

## Review focus

`document/chunking.go`: `advance` moves the cursor by exact emitted tokens; `tokenFit` binary-searches only for tokenizers that declare monotonic prefix counts.

## Known limit

The `Tokenizer` contract wants a contiguous rune partition. BPE and SentencePiece adapters may need to coalesce tokens. No real tokenizer ships yet.

Refs #176


wesm:
Follow-up after reviewing the rebased branch against the rest of the stack. Two commits were pushed and the description is rewritten to match. What changed relative to the original head:

1. **Policy ownership.** `InputPolicy` no longer declares its own chunk vocabulary. It is derived from the profile's `EmbeddingChunkPolicyV1` through `NewInputPolicy`, which also refuses a tokenizer whose name and revision differ from the declaration. To make that check possible the chunk policy gained `tokenizer_revision`, `truncation_policy` became a closed vocabulary, and bindings gained `max_input_tokens` so the chunk budget stops doubling as the provider limit. Config carries the matching keys.
2. **Limits out of identity.** Aggregate ceilings, fitting work budgets, and the max input count moved to `GenerationLimits` and no longer enter any fingerprint. A test proves changing them leaves the generation checksum unchanged.
3. **Canonical codec.** The generation now encodes, decodes, and fingerprints through `internal/canonical`, like every other durable document contract. The hand-rolled JSON preflight scanner and the custom hash framing are gone. A persisted generation recomputes its own policy fingerprint from its fields.
4. **Removed the rune-based plan path.** `BuildEmbeddingPlan`, recipes, and distillation in `document/embedding` are deleted; they had no consumer and would have left two chunkers with unclear authority. Egress, retrieval, and eval stay.

Smaller cleanups: the built-in profiles in `NewModelInputContract` are a table, `HeadingPath` and `SourceSpan` are scalars, `chunking.go` is split into `chunking.go` and `generation.go` with a fitter struct instead of nine-parameter functions, and the unused `TruncateIndivisibleAtom` alias is removed.

Left as-is: the unrelated `trafilatura` test change is still in the branch, and the `Tokenizer` contract's fit with real BPE and SentencePiece tokenizers is recorded as a known limit in the description rather than solved here.

wesm:
Addressed the roborev findings and the Windows failure.

- **Tokenizer identity check**: fixed. `BuildEmbeddingInputs` now compares the final identity read against the identity the chunk policy declares, which validation already proved, instead of against an earlier fresh read.
- **`MaxInputs` after decode**: no change. JSON cannot amplify, so `MaxEncodedBytes` already bounds every allocation the decode can make to the size of the input. Enforcing `MaxInputs` before decoding would need a second JSON scanner, which this PR deliberately removed in favour of the shared canonical codec. The count check after decode is a policy bound, not a memory bound.
- **Windows CI**: the new chunk golden files were being checked out with CRLF. They are now marked `-text` in `.gitattributes` like the other golden oracles.

wesm:
Confirmed and fixed. The fitter narrowed each input's limits by the remaining aggregate room, so under `truncate_indivisible` a final indivisible token could be cut to the leftover budget and the build would succeed with different content. That contradicted the rule that `GenerationLimits` only reject.

Inputs are now fitted against the policy limits only, and aggregate limits reject when an input's totals are added. A new test builds the same evidence with limits set exactly to the unlimited totals and asserts an identical generation, then decrements each aggregate limit and asserts an error rather than a truncated last input. The two earlier tests that asserted the narrowing behaviour are removed since they encoded the bug.

wesm:
Confirmed and fixed. `validateModelInputContractFields` now rejects a non-empty `QueryInstruction` on `custom/v1` contracts, matching the constructor, and it does so before the fingerprint check so a hand-built or decoded contract fails on the field rather than on a stale fingerprint. Added a test that takes a valid custom contract, sets the instruction, and asserts the descriptor path rejects it.

wesm:
Confirmed and fixed. Pre-tokenization now records byte offsets next to rune offsets in one pass per unit, so probes take substrings directly. The only remaining in-token scans happen when overlap moves a token's start or truncation shortens its end, both bounded by the token itself and the existing truncation search limit. `runeRange` is gone.

Measured on a single 300k-rune unit with a 64-token budget and 8-token overlap (5357 inputs): 3.40s before, 60ms after, on the same machine. The 20-second span fuzz passes on the new offsets.

wesm:
Declining this one. There is no installed base for the field shapes this finding wants preserved.

- The latest release, v0.14.0, contains neither `ProcessingProfileV1` nor the processing catalog that stores profiles. `EmbeddingChunkPolicyV1` first appeared in #183, after that tag, and `EmbeddingBindingV1.ModelInput` in #210.
- The `processing-profile/v1` golden and its fingerprints already changed in #202 and #210 on main with no migration. This stack treats the contract as unreleased and evolving until it ships.
- The only vaults that can hold a chunk policy with `"reject"` or without `tokenizer_revision` were written by development builds of main. Those are rebuilt, not migrated.

A legacy read path or a v2 contract would be scaffolding for data that no released binary ever wrote. The right time to add versioning and migration is the release that first ships this contract.

PR #212, opened 2026-08-26
[F10–E12 16/25] Make embedding generations durable and independently activatable

salmonumbrella:
Embedding generations and canonical vector sets are now durable store authority that each processing profile can activate independently. This PR is the child of #211 and has been rebuilt on current `main`.

The store derives vector-row metadata from the canonical payload, rejects payloads that diverge from their input generation, and performs the same deep catalog validation through snapshots, exports, backups, and restores. Head and failure identity includes the processing profile, so profiles that reuse a binding name cannot overwrite each other.

Version pruning removes version-owned sets and heads while leaving immutable generation and vector authority for the shared, root-aware derivative collection path. Reverse-reference indexes keep prune, purge, and GC lookups bounded as the catalog grows.

This layer intentionally stops at the authority downstream processing uses: staging and publishing embedding sets. Corpus manifests, the per-cell coverage query, the unused active-head reader, and embedding-only root aliases are omitted; downstream processing owns set-based coverage and vector-index generations.

Canonical vectors remain separate from physical indexes, so indexes stay rebuildable and non-authoritative.

Refs #176

### DIFF
--- a/document/chunking_test.go
+++ b/document/chunking_test.go
@@ -706,6 +706,37 @@ func TestEmbeddingInputGenerationDecodeAppliesCallerBounds(t *testing.T) {
 	require.ErrorContains(t, err, "decode bound is invalid")
 }
 
+func TestEmbeddingInputGenerationValidatesEvidenceSpans(t *testing.T) {
+	evidence := testChunkEvidence(t, []sourceChunkUnit{{text: "é界AB", heading: []string{"Evidence"}}})
+	for _, testCase := range []struct {
+		name   string
+		mutate func(*GeneratedEmbeddingInput)
+	}{
+		{"missing unit", func(input *GeneratedEmbeddingInput) { input.SourceSpan.UnitIndex = 1 }},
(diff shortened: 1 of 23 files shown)


wesm:
rebasing

wesm:
All merge-blocking items from the earlier review, plus the two latest RoboRev findings, are addressed:

- Version pruning removes version-owned heads and sets, but leaves retained generations and vector sets for root-aware garbage collection. Restore now verifies those retained vector sets from their own immutable metadata.
- Vector-row metadata is always derived from the canonical payload, and deep validation runs for snapshots, exports, backups, and restores.
- The unused per-cell coverage query, active-head reader, corpus catalog, and embedding-only root aliases were removed. Downstream processing uses its set-based coverage path instead.
- Reverse-reference indexes now cover prune, purge, and garbage-collection lookups.
- Heads and failures are keyed by processing profile, so profiles cannot overwrite one another.
- SQLite no longer hard-codes evolving contract, input-kind, or failure-code values. Shared Go validation still rejects unsupported values.

The broader ideas about store-derived IDs, a generic derivative ledger, and reducing JSONL volume remain follow-up design work; they were not part of the merge-blocking recommendation.

PR #213, opened 2026-08-26
[F10–E12 17/25] Add explicit local OpenAI-compatible embeddings

salmonumbrella:
## What changed

- Added a local OpenAI-compatible embedding adapter for Ollama, llama.cpp, LM Studio, and compatible operator-run endpoints through one fixed `POST /v1/embeddings` contract.
- Made the model-input contract explicit. Nomic, E5, BGE-M3, GTE, Qwen3, query-only instructions, and complete custom profiles are selected deliberately; model names never change formatting behavior.
- Pinned vector-space identity to either an operator-controlled deployment epoch or one exact provider-attested revision header. Responses must also echo the requested model and return one correctly indexed, finite vector of the declared dimension for every input.
- Bounded batches, rendered input, request and response bytes, timeouts, redirects, and credentials. Provider errors and malformed responses do not expose request text, vectors, secrets, or raw response bodies.

This PR is the child of #212 and adds the local embedding adapter on current `main`.

## Why

An OpenAI-compatible endpoint only standardizes the wire shape. It does not prove that a mutable model alias, document prefix, query instruction, dimension, or local deployment still represents the same vector space.

Making those choices explicit lets operators swap among compatible local servers without hidden model-name heuristics or silently mixing incompatible document and query vectors.

## Usage

Create an `openaiembed.Profile` with the endpoint origin, canonical embedding descriptor, explicit model-input contract, capacity bounds, and either a deployment epoch or revision header. Add a named bearer-secret binding only when the local endpoint requires one, then construct the provider with `openaiembed.New` and call it through `document.ExecuteEmbedding`.

No daemon or CLI surface is added in this layer; worker integration follows in E7.

Refs #176

### DIFF
--- a/document/openaiembed/client.go
+++ b/document/openaiembed/client.go
@@ -1,0 +1,596 @@
+// Package openaiembed implements one bounded OpenAI-compatible text embedding
+// endpoint for operator-controlled local deployments.
+package openaiembed
+
+import (
+	"bytes"
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	json "encoding/json/v2"
+	"errors"
+	"fmt"
+	"io"
+	"math"
+	"mime"
+	"net"
+	"net/http"
+	"net/netip"
+	"net/url"
+	"reflect"
+	"slices"
+	"strconv"
+	"strings"
+	"time"
+	"unicode"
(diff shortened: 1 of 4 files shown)


wesm:
rebasing

PR #214, opened 2026-08-26
[F10–E12 18/25] Add bounded hosted Voyage and Mistral embeddings

salmonumbrella:
Docbank can export and evaluate hosted embeddings through Voyage 4 text, Voyage contextual embeddings, Voyage direct-file inputs, and Mistral Embed. These hosted profiles remain export-only: a mutable model alias does not establish an immutable vector space suitable for retrieval activation.

Each profile requires HTTPS and pins its endpoint, network policy, input formatting, credential binding, and execution limits. Responses must match the expected model, dimensions, indices, and vector contract. Contextual invocations contain one document's ordered chunks so unrelated documents do not share context. Direct-file execution retains capability checks and the core's upload ownership and verification.

Construct a `voyage.EmbeddingProfile` or `mistral.EmbeddingProfile`, create its provider with `NewEmbeddingProvider`, and call it through `document.ExecuteEmbedding`. This layer adds no daemon or CLI surface.

Parent: #213 (merged). This PR contains only the hosted embedding changes on current `main`.

Refs #176

PR #215, opened 2026-08-26
[F10–E12 19/25] Add a deterministic rebuildable vector index

salmonumbrella:
Docbank can find the nearest embedding vectors with exact, repeatable results and rebuild the search index from stored vector sets. Cosine, dot-product, and Euclidean searches break ties consistently.

An open index keeps one copy of the vectors. Saved bytes are produced only when requested. On a seeded 10,000 × 128 corpus, building and saving the index allocated 39.9 MB, down from 91.5 MB. An opened index retained about 7.1 MB.

Opening checks file integrity and verifies every vector against its recorded source set. Each source set must be included in full. The binary format remains unchanged. Search scans and sorts all rows, so query cost grows with index size.

Daemon startup and stop can also recover when a recorded daemon is still running but no longer answers requests. Replacement waits for that process to exit.

Parent: #214 (merged). Refs #176.


wesm:
rebasing

wesm:
giving this a review pass after rebasing

PR #216, opened 2026-08-26
[F10–E12 20/25] Publish embedding sets independently

salmonumbrella:
Docbank runs queued embedding jobs and publishes each binding independently, so a provider failure leaves other completed sets usable. The daemon supports OpenAI-compatible chunk embeddings and Voyage original-file embeddings with named credential bindings and validated runtime descriptors.

Jobs remain disposable runtime state. After restore, the worker rebuilds them from retained input generations and fresh consent. Publication uses the existing profile-scoped catalog, source and attachment checks, monotonic fencing tokens, and purge rules. Claims select only runtimes configured in the daemon.

Chunk input is hydrated once per attempt; original files are reopened for every provider retry. The worker does not create input generations or apply profiles to new documents.

Parent: #215 (merged). Refs #176.

### DIFF
--- a/cmd/docbank/daemon.go
+++ b/cmd/docbank/daemon.go
@@ -52,6 +52,36 @@ var daemonRunCmd = &cobra.Command{
 	},
 }
 
+type embeddingRuntimeReadiness interface{ Ready() bool }
+
+type embeddingJobRunner interface {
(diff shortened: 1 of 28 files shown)


wesm:
rebasing

wesm:
invalid

PR #217, opened 2026-08-26
[F10–E12 21/25] Publish rebuildable vector indexes

salmonumbrella:
Docbank now builds local vector indexes from published embedding sets, validates each candidate before activation, and keeps earlier generations available while readers hold leases. Indexes remain disposable: snapshot restore rebuilds retained vectors without contacting an embedding provider.

The worker reads through the managed blob store and respects the current catalog's profile and source eligibility. Shared vectors are indexed once across profiles, while each distinct source membership gets its own generation identity. New membership supersedes obsolete build claims, failed attempts release their claims, and empty membership retires the active index without disrupting leased readers.

Idle scans check publication metadata without reloading index bytes. Payload reads and index construction leave maintenance admission free; catalog publication still checks the build claim and current source. Persistent catalog failures appear in job status, and transient SQLite contention and blob-store outages receive bounded retries. Cleanup preserves candidates with live build claims, and new readers must acquire an index that matches the current source.

The worker records missing vector payloads as unavailable coverage and keeps running, so restored payloads or new embeddings can recover that coverage. Corrupt payloads remain errors. Full backup restore still requires a complete snapshot; this change does not permit omitted required attachments.

The vector-index layout uses storage schema version 6. Released vaults retain their existing JSONL upgrade path; unshipped schema versions 4 and 5 have no on-disk adapter.

Parent: #216 (merged). Refs #176.


wesm:
rebasing

Issue #197, opened 2026-08-24
add: glob-based include/exclude and cloud-placeholder awareness for mixed trees


Joi:
I am starting to use docbank as the system of record for the documents scattered across my Obsidian vault and two Google Drive accounts (Drive for Desktop mounts on macOS). Two things came up on the first ingest that `docbank add` cannot express today.

## 1. Documents-only ingest of a mixed tree

`--exclude` matches an entry name or a relative path, not a pattern. A real tree is mostly things I do not want in the vault: my larger Drive preflights at 15,525 files / 3.2 TB logical, of which 1,958 `.wav`, 738 `.zip`, 365 `.mp4` and 2,491 `.jpg` account for almost all the bytes, while the ~2,000 PDFs, Office files and Keynote packages I actually want are 11 GB. The vault side is the same shape: 28k markdown notes already versioned elsewhere, ~1,500 PDFs/DOCX/PPTX I want.

Excluding 13,000 media files by name is not practical, so I wrote a small wrapper that walks the tree, filters by extension, groups by directory and calls `docbank add <files...> --dest <mirrored path>` per directory. It works (idempotent by hash, as documented), but it means the CLI's own preflight/progress/aggregate report is lost and the provenance `source_description` becomes a list of files instead of the tree.

Suggestion: `--include <glob>` / `--exclude <glob>` (gitignore-style, evaluated against the relative path), or a `--types pdf,docx,...` shortcut. Preflight already reports `file_types`, so the data to make that decision is right there.

## 2. Cloud-drive placeholders hydrate at network speed

Ingesting from a Drive for Desktop mount is correct but slow: the daemon streams each placeholder at the drive's download rate (~2.3 MiB/s here), and a Keynote `.key` package expands into hundreds of small files. `--preflight` is fast because it only stats. It would help if preflight could flag "this source is a cloud placeholder / dataless file" (macOS reports `SF_DATALESS` via `stat`, `chflags`) so the operator knows the ingest will hydrate, and if `add` could be told a byte budget or file-size ceiling per run (`--max-file-size`) rather than only the fixed 16 MiB text-extraction limit.

Both are quality-of-life; nothing is incorrect. Happy to test a branch. docbank 0.14.0, macOS 26.5 arm64.

Joi:
Concrete data point for the placeholder half of this. With the daemon launched by launchd (`docbank daemon run` as a LaunchAgent, background process type), server-side ingest of a Drive for Desktop placeholder fails with:

```
[...]

PR #219, opened 2026-08-26
Report cloud placeholders in preflight and explain hydration failures

Joi:
Closes half of #197 (the placeholder half; the glob include/exclude request there is untouched).

Preflight already refuses to open cloud-provider placeholders, which is the right call — but it also said nothing about how many of them a source contains, so a tree that looks like a quick import can turn into hours of provider downloads. Preflight now counts them:

```
files: 15525  directories: 2052  logical size: 2.9 TiB
pack eligible: 13043 file(s), 36.9 GiB
loose only: 2374 file(s), 1.9 TiB
rejected: 108 file(s), 1.0 TiB
cloud placeholders (not present locally): 9871 file(s), 2.4 TiB — ingest downloads these through the provider, or fails those it will not serve to the daemon
```

The count appears in the CLI report, `--json`, and `IngestPreflightReport` (`cloud_placeholders`). Placeholders stay counted in their size classes too: the report describes the same file from two angles.

The second half is the error message. When a provider declines to hydrate a placeholder for the process opening it, macOS returns `EDEADLK` and the ingest failure read as:

```
reading /Users/.../My Drive/Talks/JoiIto.pdf: resource deadlock avoided
```

which is hard to act on. It now names the cause and the fix (open the file once from a user session, or mark it available offline, then retry). In my case the daemon runs as a launchd background job and the provider serves an interactive session but not that daemon — the same file imported fine after one `cat` from a shell.

Detection is `SF_DATALESS` on darwin; other platforms compile to no-ops and report zero, with tests on both paths.

Found while running docbank as the document system of record over a vault and two Google Drive accounts (~11.5k document nodes). Happy to adjust the wording or the field name.

### DIFF
--- a/cmd/docbank/add.go
+++ b/cmd/docbank/add.go
@@ -142,6 +142,11 @@ func printIngestPreflight(cmd *cobra.Command, rep api.IngestPreflightReport) {
 		rep.LooseOnly.Files, formatBackupBytes(rep.LooseOnly.Bytes))
 	_, _ = fmt.Fprintf(out, "rejected: %d file(s), %s\n",
 		rep.Rejected.Files, formatBackupBytes(rep.Rejected.Bytes))
+	if rep.CloudPlaceholders.Files > 0 {
+		_, _ = fmt.Fprintf(out, "cloud placeholders (not present locally): %d file(s), %s — "+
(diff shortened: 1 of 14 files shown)


wesm:
Thank you for the fix! Will merge this pending CI and release it in the near future (but you are welcome to run off main and have the fix sooner)

PR #220, opened 2026-08-26
[F10–E12 22/25] Add explained hybrid document search

salmonumbrella:
Docbank can run lexical, semantic, or hybrid document search and explain where each result came from. Omitted mode and `auto` stay lexical; only explicit `semantic` or `hybrid` requests send query text to an embedding provider.

Semantic search uses the query contract recorded with the active vector space and returns candidates only from the requested processing profile, binding, and input kind. Chunk matches reduce to one result per document, and hybrid search combines lexical and semantic ranks with deterministic reciprocal-rank fusion.

Results include bounded lexical excerpts, stable rendition or embedding evidence, coverage, the actual mode used, and a compact trace. Explicit semantic searches report provider, catalog, index, and lease failures rather than silently changing modes.

Semantic search currently scans all indexed vectors for exact results; it resolves paths only for the selected documents.

Parent: #217 (merged). Refs #176.

### DIFF
--- a/internal/processing/embedding_worker_test.go
+++ b/internal/processing/embedding_worker_test.go
@@ -1132,7 +1132,7 @@ func (embeddingIntegrationTokenizer) Tokenize(text string, limit int) ([]documen
 	return result, nil
 }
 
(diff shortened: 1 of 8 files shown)


wesm:
rebasing

PR #221, opened 2026-08-26
[F10–E12 23/25] Add optional expansion and reranking to search

salmonumbrella:
## What changed

Search can now add separately configured query expansion and reranking stages without making either one part of the default retrieval path. Expansion returns a small bounded set of variants; reranking can only reorder the already-authorized candidate set and receives bounded excerpts and provenance rather than document bodies.

Each stage has its own profile, authorization, deadline, failure policy, named degradation, and sanitized receipt. Search scope is validated and canonicalized before provider authorization. Both expanded searches and reranking-only searches recheck candidates against current scope, node revision, rendition or embedding evidence, vector source, and final semantic coverage. Revalidation refreshes current paths and discards invalid candidates before reranking.

Parent: #220 (merged). This PR contains only the optional expansion and reranking changes.

## Why

Expansion and reranking are useful only when they cannot quietly widen access, send unbounded content, or turn stale evidence into a current result. Keeping them optional and independently authorized makes the extra quality/latency/privacy tradeoff explicit, while the base lexical, semantic, hybrid, and automatic modes remain available with both stages disabled.

## Usage

Add an `ExpansionConfig`, a `RerankingConfig`, or both when constructing the retrieval `Searcher`. Each enabled stage needs its own typed provider, authorizer, profile limits, deadline, and `degrade` or `fail_closed` policy. Successful authorization must return a non-nil `ProviderEgressLease`, which retrieval holds through provider completion and closes on success, failure, or cancellation. Consent revocation must acquire the corresponding exclusive fence; the store’s `BeginProviderEgress` helper coordinates with its existing revocation lock. Leave both configs disabled to keep expansion and reranking disabled.

Refs #176

### DIFF
--- a/internal/retrieval/egress_test.go
+++ b/internal/retrieval/egress_test.go
@@ -1,0 +1,159 @@
+package retrieval
+
+import (
+	"context"
+	"errors"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"go.kenn.io/docbank/internal/store"
+)
+
+func TestProviderStagesHoldConsentThroughExecution(t *testing.T) {
+	for _, stage := range []ProviderStage{ProviderStageExpansion, ProviderStageReranking} {
+		for _, outcome := range []string{"success", "failure", "canceled"} {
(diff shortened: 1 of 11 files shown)


wesm:
rebasing

PR #223, opened 2026-08-26
[F10–E12 24/25] Add a standard embedding provider bridge

salmonumbrella:
## What changed

- Adds one fixed, synchronous `docbank-embedding/v1` endpoint for document text, query text, and exact authorized original-file uploads.
- Publishes strict OpenAPI and JSON Schema contracts, including deterministic request manifests and checksums that bind the provider, vector space, authorization, inputs, headings, and source spans.
- Binds origin, egress, authentication, model-input policy, and execution limits into the provider identity. Named bearer credentials are supported, while no-auth profiles are limited to an explicit operator-network deployment.
- Rejects redirects, schema drift, oversized or malformed responses, non-finite or wrong-size vectors, and changed source bytes. Provider and transport failures stay classified without exposing request text, files, credentials, or vectors.
- Parent: #221 (merged). This PR contains only the standard embedding bridge.

## Why

Hosted and self-hosted providers currently need a bespoke adapter inside Docbank. This gives them one bounded protocol to implement without exposing arbitrary request templates or weakening the vector-space and source-version guarantees that retrieval depends on.

## Usage

Implement `POST /docbank-embedding/v1/embeddings` from the shipped OpenAPI and JSON Schema contracts. Configure a canonical embedding descriptor, the exact service origin and egress policy, execution limits, and either a named credential binding or an explicit operator-network no-auth profile, then call the client through Docbank's embedding execution boundary.

Refs #176

### DIFF
--- a/cmd/docbank/daemon_test.go
+++ b/cmd/docbank/daemon_test.go
@@ -109,6 +109,8 @@ func TestServeRecoversInterruptedRestoreBeforeInitializingVault(t *testing.T) {
 		}
 	})
 	healthClient := &http.Client{Timeout: time.Second}
+	// Recovery and fresh SQLite initialization can exceed ten seconds on
+	// Windows CI. This checks recovery ordering, not startup performance.
 	require.Eventually(t, func() bool {
 		records, listErr := client.RuntimeStore(dir).List()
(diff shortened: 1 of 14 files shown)


wesm:
rebasing

PR #224, opened 2026-08-26
[F10–E12 25/25] Add fixed hosted OpenAI embeddings

salmonumbrella:
Go applications can generate document and query embeddings through a dedicated hosted OpenAI adapter. It calls only `https://api.openai.com/v1/embeddings`, uses `text-embedding-3-large`, and requires an explicit dimension from 1 to 3,072. The model-input policy and an operator-controlled compatibility epoch are part of vector-space identity.

Profiles require a named API-key resolver and sealed `api.openai.com:443` egress. Requests have independent item, batch, input, request, response, and timeout limits. Responses must contain complete unique indices, the expected model and usage fields, and finite unit-length vectors of the configured dimension. Unknown response members remain rejected as part of the strict hosted contract.

Errors distinguish unavailable or rejected credentials, transient failures, capacity limits, and permanent failures without exposing provider bodies or secrets. Adapter timeouts are transient; caller cancellation stays identifiable. Egress-policy failures and oversized individual inputs are permanent, while an authorization that exceeds the profile is rejected locally.

Use `openaihosted.New` with a canonical profile, then call `document.ExecuteEmbedding`. Zero limits select a 128-item batch and a 64 MiB response budget. Larger batches or dimensions may need a larger response budget, which measures encoded JSON including number text and metadata.

This PR adds the public Go package. CLI/daemon configuration and worker error-classifier registration remain follow-up work; the daemon currently dispatches only to the OpenAI-compatible and Voyage adapters.

Parent: #223 (merged). This PR contains only the hosted OpenAI adapter.

Refs #176

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,7 +170,7 @@ jobs:
           $ErrorActionPreference = 'Stop'
           $zip = Join-Path $env:RUNNER_TEMP 'llvm-mingw.zip'
           $url = "https://github.com/mstorsjo/llvm-mingw/releases/download/$env:LLVM_MINGW_VERSION/$env:LLVM_MINGW_NAME.zip"
-          Invoke-WebRequest -Uri $url -OutFile $zip
+          Invoke-WebRequest -Uri $url -OutFile $zip -MaximumRetryCount 3 -RetryIntervalSec 5
           $actual = (Get-FileHash -Path $zip -Algorithm SHA256).Hash.ToLower()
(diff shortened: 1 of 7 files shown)


wesm:
rebasing

mariusvniekerk:
For all this embedding work should we do this stuff in `kenn-io/kit` instead so that we can more easily reuse it across the rest of the applications.  Doing it here is rather duplicative in work?

wesm:
You're welcome to contribute some pull requests to help

PR #225, opened 2026-08-26
[E13–S9 01/15] Add Cohere embedding and reranking

salmonumbrella:
Go applications can generate text and authorized image embeddings with Cohere Embed v4 through `document/cohere`. The adapter supports 256, 512, 1024, and 1536 dimensions, with document and query inputs bound to an explicit vector-space identity. Construct it with `cohere.New` and execute it through `document.ExecuteEmbedding`.

The separate `internal/retrieval/cohere` adapter implements Cohere Rerank v4 pro and fast. It sends bounded query text and candidate excerpts, maps returned positions to local candidates, and keeps excerpt byte budgets independent of the provider's token truncation limit. This PR adds the adapter packages; CLI/daemon configuration and retrieval runtime wiring are not included.

Embedding and reranking require independent profiles and named credentials. Both use sealed egress and bounded requests, responses, timeouts, and retry guidance. Shared Cohere protocol code handles response metadata and HTTP errors, while each adapter retains its own error identities. Receipts use billed token counts; total token counts are validated separately. Image uploads are verified against their authorized bytes and media identity, while response dimensions may reflect provider processing. Cancellation interrupts blocked upload reads.

Parent: #224 (merged).

Refs #176


wesm:
rebasing

PR #227, opened 2026-08-28
[E13–S9 03/15] Add ZeroEntropy embedding and reranking

salmonumbrella:
Go applications can generate document and query embeddings with ZeroEntropy `zembed-1` through `document/zeroentropy`. Profiles support 40, 80, 160, 320, 640, 1280, and 2560 dimensions, float or base64 encoding, and auto, fast, or slow latency. Document and query inputs are sent separately and results retain the input order.

The separate `internal/retrieval/zeroentropy` adapter adds `zerank-2` reranking. It sends bounded queries and candidate excerpts, requires a complete set of finite scores, and maps provider indices back to local document identities. Both adapters use independent profiles, named credentials, fixed ZeroEntropy egress, and bounded requests and responses. Optional receipts record usage and latency settings without document text.

This PR adds the adapter packages. CLI/daemon configuration and runtime wiring are not included.

Parent: #226 (merged).

Refs #176

PR #228, opened 2026-08-28
[E13–S9 04/15] Add self-hosted BGE-M3 and Qwen3 profiles

salmonumbrella:
## What changed

- Adds reviewed dense-only BGE-M3 and Qwen3 profiles to the existing operator-controlled OpenAI-compatible adapter.
- Pins immutable model and tokenizer commits, pooling, context window, native dimension, cosine metric, unit-length normalization, and query/document formatting into vector-space identity.
- Keeps BGE sparse and multi-vector output, plus Qwen dimension transforms, outside `vector-set/v1`; caller-owned egress policy storage is also isolated before use.

## Why

A self-hosted endpoint exposes a generic API, but that API does not prove how the deployment produces its vectors. These profiles make the hidden model-side contract durable, so changing weights, tokenization, pooling, or output shape cannot silently mix incompatible vector spaces.

## Usage

Use `openaicompat.BGEM3Profile` for `BAAI/bge-m3`, or `openaicompat.Qwen3Profile` with the 0.6B, 4B, or 8B model and a required retrieval instruction. Supply the exact 40-character model and tokenizer Git revisions plus the operator deployment epoch. BGE-M3 uses its native 1024 dimensions; Qwen3 uses 1024, 2560, or 4096 dimensions by model size. Both profiles are dense-only.

Refs #176

Parent: #227 (merged).

### DIFF
--- a/document/openaicompat/client.go
+++ b/document/openaicompat/client.go
@@ -73,6 +73,7 @@ type Profile struct {
 	Origin                 string
 	Descriptor             document.EmbeddingDescriptor
 	ModelInput             document.ModelInputContract
+	DeploymentContract     *DeploymentContract
 	SecretBinding          string
 	DeploymentEpoch        string
 	ProviderRevisionHeader string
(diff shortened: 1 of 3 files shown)


wesm:
rebasing

PR #229, opened 2026-08-28
[E13–S9 05/15] Export sanitized Markdown as disposable QMD collections

salmonumbrella:
This adds an internal API for exporting active sanitized Markdown as disposable QMD collections. Opaque, deterministic paths keep original filenames out of the index, and a manifest maps each entry to its exact document version and rendition. CLI and daemon wiring remains outside this PR.

`qmdexport.PublishActive` streams documents into staging and verifies their size, UTF-8, and checksums before `CURRENT` selects a complete generation. Existing export paths must have private permissions; new files and directories are restricted before content is written. Each active profile gets its own entry, so a document with multiple profiles can appear more than once.

Concurrent exports take their snapshots in publication order, so a delayed run cannot replace a newer document list. Published generations remain on disk for existing readers. Callers must coordinate their removal once consumers finish; publication cleans only abandoned staging directories. The collection is rebuildable from Docbank and does not become document authority.

Refs #176. Parent: #228 (merged).


wesm:
rebasing

PR #230, opened 2026-08-28
[E13–S9 06/15] Use operator-hosted QMD without trusting its index

salmonumbrella:
Search an operator-hosted QMD collection and return only results that still match current Docbank documents. The internal bridge checks each result against the active export manifest, document version, and rendition. Stale entries or a generation change during the query cause the request to fail.

Scoped searches fetch candidates within a configured limit, remove out-of-scope hits, and keep the highest-ranked hit for each document across active rendition profiles. The result reports when limits may hide additional matches, including when no scoped matches were found.

Queries hold a consent lease through the HTTP request and response, use a named bearer credential, and apply configured network limits and deadlines. The bridge sends query text and search controls; it does not upload original files or retained Markdown. CLI and daemon wiring remain outside this PR.

Refs #176. Parent: #229 (merged).

### DIFF
--- a/internal/qmdexport/export.go
+++ b/internal/qmdexport/export.go
@@ -71,6 +71,7 @@ type Options struct {
 	MaxDocuments     int
 	MaxDocumentBytes int64
 	MaxTotalBytes    int64
+	MaxManifestBytes int64
 }
 
 type publishHooks struct {
@@ -234,6 +235,9 @@ func publish(ctx context.Context, root, collection string, sources []Source, cat
 	if err != nil {
 		return Receipt{}, fmt.Errorf("encode qmd export manifest: %w", err)
 	}
+	if int64(len(manifestBytes)+1) > bounds.MaxManifestBytes {
+		return Receipt{}, errors.New("qmd export manifest exceeds bound")
+	}
 	manifestBytes = append(manifestBytes, '\n')
 	if _, err := writePrivateFile(filepath.Join(stage, "manifest.json"), bytes.NewReader(manifestBytes)); err != nil {
 		return Receipt{}, err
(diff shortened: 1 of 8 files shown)


wesm:
rebasing

wesm:
## Review of PR #230

Pushed `9f34838` with the fixes below. The three findings from the earlier roborev review (consent held through egress, scope-aware over-fetch with truncation, multi-profile dedup) are all addressed in `2c7cf41` and covered by tests that run against a real store.

**Findings: 1 P2, 3 P3, 4 P4.** Fixed 4 (all P2/P3), dismissed 4 (P4, informational).

### Fixed

- **P2 (performance)** — `Search` parsed the full export manifest from disk three times per query (`LoadCurrent` before mapping and twice more as generation fences). A manifest can hold 100k entries and run to tens of MB. Since a generation ID *is* the manifest checksum and the loader rejects any manifest whose checksum differs, the fences now read only the 65-byte `CURRENT` pointer via new `qmdexport.CurrentGeneration`. Both fence points are kept so a mid-query publish still reports `ErrStaleGeneration` rather than stale authority.
- **P3 (style)** — `Search` was 114 lines, over the 100-line limit. The HTTP request/response handling moved to `client.exchange`; `Search` is now 83 lines with no behavior change.
- **P3 (correctness)** — a JSON marshal failure was reported as "request exceeds bound". It now reports an encoding failure.
- **P3 (tests)** — no coverage for request rejection before egress (bad search type, blank query, zero/over-cap limit, out-of-range score, query+intent byte budget) or for non-200 / non-JSON responses. Added two tests; verified they fail when the corresponding guards are removed.

### Dismissed (P4)

- `RevalidateQMDExportCandidates` issues two queries per candidate (up to 1,000 → 2,000 queries in one transaction). Bounded, local SQLite, single snapshot; not worth a batched query until it shows up in profiling.
- `RevalidateQMDExportCandidates` re-normalizes scope options the client already normalized. Cheap and correct at the store boundary; the interface takes raw options.
- Filesystem-root check exists in two forms (`filesystemRoot` in profile.go, `filepath.Dir(abs) == abs` in load.go). Both are correct and volume-aware; cosmetic.
- `info.Size() < 0` in `readRegularBounded` is unreachable. Harmless.

### Verification

`make lint` 0 issues, `go test -tags fts5 ./...` green, touched packages also green with `CGO_ENABLED=0`, `prek run` passes.

PR #231, opened 2026-08-28
[E13–S9 07/15] Expose document processing through the embedded API

salmonumbrella:
Go applications can now process and search documents through the embedded `Vault` API. Callers preview what a provider will receive and what the vault will retain, then approve the selected document version and any disclosed filename.

Renditions require current, non-trashed sources and recheck disclosed filenames, including on resume. Equivalent requests share durable work. Processing waits through retries, supports cancellation, and reports terminal failures. Rejected requests stop before dependent embeddings run; optional provider failures leave partial results available.

Search stays within the selected versions and vault. Coverage marks superseded or deleted sources stale instead of counting them as complete. Semantic and hybrid queries require consent; automatic search stays lexical.

Retained Markdown includes provenance and navigation metadata, while search text and embedding inputs use the document body. Each upload directory has one vault or daemon owner; reopening clears abandoned copies.

Refs #176. Parent: #230 (merged).


wesm:
rebasing

wesm:
## Review follow-up (`03c8eaf`)

Findings from the combined review: 2 High, 3 Medium. Fixed 3, dismissed 2, plus one small cleanup from my own pass.

### Fixed

- **Frontmatter carried raw HTML from provider-controlled headings** (`document/rendition_frontmatter.go`). Confirmed: navigation titles came from `HeadingPath` / `Locator.Name` through `strconv.Quote` only, so `<script>` survived into the retained artifact while the body was sanitized. Frontmatter strings now write `<`, `>`, `&` as YAML `\u` escapes. A regression test parses the header with `yaml.v3` and checks the decoded title is exact while the raw bytes contain no markup.
- **Provider runtimes silently deduplicated by descriptor fingerprint** (`internal/processing/service.go`). Confirmed: the fingerprint hashes model identity only, not endpoint or credentials, so a second profile's distinct provider would have executed through the first profile's provider. Opening a vault now fails when two profiles supply different provider values for one fingerprint; sharing the same value is still allowed. Test added in `vault_external_test.go`.
- **Abandoned embedding jobs reported as completed** (`internal/processing/service.go`). Confirmed by reading `aggregateStatus`. It now reports `abandoned` explicitly and only says `completed` when every embedding job is completed. Test added.

### Dismissed

- **Envelope makes staged rendition fail the deterministic rebuild check** (`rendition_worker.go:1027`). The deep-compare lives in `ArtifactPublisher.PublishRendition` (`artifacts.go:213`), which has no production callers; only tests use it. The worker publishes through `validateStagedRendition` + `StageRenditionJobBuild` + `PublishRenditionJob`, which checks the Markdown checksum against the enveloped bytes and passes. The embedded end-to-end tests process and read an enveloped rendition successfully. Worth noting separately: `ArtifactPublisher` is dead production code and its graph validator now disagrees with the worker's candidate builder.
- **`failure_code` column added without a schema version bump** (`schema.sql:1056`). `rendition_job_waiters` does not exist in any released tag (v0.14.0 is at schema version 3; current is 6, unreleased). AGENTS.md rules out migration ladders for layouts that never shipped, and every other unreleased schema change on `main` since v0.14.0 (nine commits) has landed the same way. No released vault is affected.

### Cleanup

- The public `MaxDocumentSourceFenceIDs`, `DefaultDocumentSearchLimit`, `MaxDocumentSearchLimit`, and `MaxRenditionBytes` constants now alias the internal values instead of restating them.

`make lint` is clean; `go test -tags fts5 ./...` passes in CGO mode, and the touched packages pass with `CGO_ENABLED=0`.

PR #232, opened 2026-08-28
[E13–S9 08/15] Expose document processing through the authenticated HTTP API

salmonumbrella:
Applications can process documents through authenticated HTTP and a typed Go client, then read verified renditions and search selected versions. Before approving work, callers can review the exact document version, any disclosed filename, and whether another consent grant is needed.

Purge previews cover only selected derivatives, so unrelated vault activity does not invalidate approval. Abandoned work and cleanup from earlier purges remain with ordinary garbage collection. If cleanup fails after live derivatives are removed, callers receive the committed receipt and a partial or deferred result. Originals and immutable backup copies remain intact.

Processing shares the embedded API's source checks, cancellation, and backup coordination. The daemon advertises only profiles it can execute; its current runtime bundle supports direct-file embeddings.

Refs #176. Parent: #231 (merged).


wesm:
rebasing

PR #233, opened 2026-08-28
[E13–S9 09/15] Add daemon-first document processing commands

salmonumbrella:
Preview and run document processing, read verified Markdown renditions, and search selected document versions from the CLI. All commands use the authenticated daemon.

`docbank processing profiles|plan|build|status` lists available profiles, previews what providers receive, and reports progress or partial failures. Plans show the exact document version and any disclosed filename. Builds require the reviewed plan fingerprint and explicit `--consent`. JSON output supports scripts; `build --ndjson` emits one job record and one terminal status record.

`docbank rendition get` verifies the complete artifact before writing to stdout. `docbank search --mode lexical|semantic|hybrid|auto` searches only the versions supplied through `--source-version`, using a named profile. For semantic or hybrid search, profiles with multiple embedding bindings require `--binding`. Auto and lexical search do not use an embedding binding. Existing search behavior, including filter-only queries, remains available when `--mode` is omitted.

Refs #176. Parent: #232 (merged).


wesm:
rebasing

PR #234, opened 2026-08-28
[E13–S9 10/15] Add web document processing and rendition views

salmonumbrella:
Process a selected document version in the web app, review what providers receive, and read its verified Markdown rendition.

Choose **Process and retrieve** to select a profile and review disclosure, consent, and retained artifacts. Rendition and embedding coverage appear separately, so a rendition stays readable while optional embeddings are unavailable. Search stays within the selected version and shows the mode used and any fallback.

Accepted jobs remain visible if their response is interrupted. Stopping the daemon cancels active processing and waits for cleanup before closing the vault. Required processing failures return a CLI error; a failed status read reports that status is unavailable. **Revoke all processing consent** revokes this operator's grants across all profiles.

The rendition viewer verifies content and shows completeness, warnings, and document navigation.

Reviewed provider plan and consent revocation

Completed rendition with optional embeddings unavailable

Verified Markdown rendition

Screenshots use the real daemon and a temporary synthetic vault.

Refs #176. Parent: #233 (merged).

PR #235, opened 2026-08-28
[E13–S9 11/15] Add document processing and rendition views to the TUI

salmonumbrella:
Review and run processing for the selected document version from the TUI. Press `P` to choose a profile and review provider disclosures, filenames, backup effects, and separate rendition and embedding coverage. `b` runs the reviewed plan.

When consent is needed, confirmation grants ongoing permission for that profile configuration and the vault's daemon operator across documents and searches. It has no expiry. The web app or API can revoke all processing consent.

Use `R` to read verified Markdown (up to 256 KiB), or `/` to search only the selected version. Run failures keep coverage visible, completed runs refresh it, and late replies cannot replace newer results or job status. A new run clears the previous job; an accepted job remains available after an interrupted response. Rendition reads check the selected version, and filenames and document text display without executing embedded terminal commands.

For a 256 KiB rendition, the scroll-and-render benchmark fell from 39.3 ms to 0.55 ms per update (median of three runs).

TUI processing consent review

Rendered by the TUI using synthetic test data. The scope and confirmation controls also fit at 80×24 with a two-step provider flow.

Refs #176. Parent: #234 (merged).


wesm:
rebasing

PR #236, opened 2026-08-28
[E13–S9 12/15] Prove source-fenced document retrieval

salmonumbrella:
Document search now scores only vectors from the caller’s allowed source versions. Docbank selects eligible rows from current, live attachments in the same transaction as search coverage, while keeping the processing profile and binding checks. Missing row authority or unknown rows stop the search.

A synthetic consumer test follows the source fence through lexical search, query expansion, vector search, and reranking. The consumer retains only the stable document identity and its allowed versions, then checks visibility again before presentation so a later revocation suppresses the result.

Refs #176. Parent: #235 (merged).


wesm:
rebasing

PR #238, opened 2026-08-28
[E13–S9 14/15] Show private processing disclosures, progress, and coverage

salmonumbrella:
People using the CLI, TUI, or web app can review who receives their documents and what is retained before granting consent. They receive a durable job ID as soon as work is accepted, and later progress stays tied to the reviewed document and profile. Coverage distinguishes rebuilding from missing required results and shows whether a previous result remains available.

Rendition reads verify the retained content within a 64 MiB limit and work with ordinary browser streams. Search results retain their exact source identities, including long paths. Embedded Go applications must disclose each network provider’s endpoint when configuring processing; the updated guide shows how.

`scripts/test-private-processing.sh` checks consent, retained Markdown, and lexical, semantic, and hybrid search using a disposable synthetic vault. It requires explicit private IP destinations and allowed CIDRs. Its network report covers Docbank’s provider connections; operators must check provider onward traffic separately.

Web processing review with synthetic data

TUI processing review with synthetic data

Refs #176. Parent: #237 (merged).

PR #239, opened 2026-08-29
[E13–S9 15/15] Add a bounded daemon-first MCP server

salmonumbrella:
Local MCP clients can browse and search the vault, inspect exact document versions, and read bounded Markdown renditions with `docbank mcp`. Requests use the authenticated local daemon and retain its ownership and source-version checks. Clients must support protocol `2026-07-28`.

Stdio is the default. Optional loopback HTTP requires a separate bearer credential and bounds request size, response size, concurrency, and deadlines. Paging does not consume a shared daemon cursor quota. Cancellation interrupts blocked I/O, and recovery never replays daemon reads that already received a response.

Nine read tools are enabled by default. `--allow-processing` adds an enqueue tool requiring the exact plan reviewed in the same MCP process and existing operator consent. It cannot grant consent or automatically replay an ambiguous start.

Refs #176. Parent: #238 (merged).


wesm:
rebasing

PR #241, opened 2026-08-30
docs: make Docbank's site and guides easier to use

wesm:
Give readers one place to learn, use, and integrate Docbank: a product page at `/`, a document guide at `/guide/`, and operating documentation at `/docs/`. The guides lead with reader tasks, use plain language, and link to the reference that owns each contract. Every public page has a Markdown companion.

The documentation now covers the functionality added since v0.14.0, including saved queries and highlight sets, tag groups and colors, embedded recovery, source metadata, previews, and the expanded processing-provider packages. It distinguishes public interfaces from internal components: current CLI, web, and TUI search remains lexical, while hybrid retrieval and QMD are internal integrations. Saving a query definition does not execute it.

The site builds and previews all three tiers together from an explicit public allowlist. Builds consume an exact reviewed screenshot commit. Before uploading, the production wrapper checks Vercel's file report against the same allowlist and size limit used by CI. Production promotion still requires a release-eligible source and maintainer authorization; deployment code comes from trusted merged source. This PR does not deploy the website.

Opening an existing vault also avoids repeated disk syncs when rebuilding the temporary schema used for validation. Creating that schema in one transaction reduced an isolated pure-Go check from 171 syncs to 9, without changing the vault layout or the schema checks.

The refreshed screenshots come from the real application using the repository's temporary synthetic vault:

The current tag filter groups related tag names and keeps stable tag colors.


wesm:
@salmonumbrella i'm grinding on this some, would appreciate your feedback after I push the incoming iteration

salmonumbrella:
@wesm tag me for feedback when you're ready

PR #249, opened 2026-09-01
Add durable visual preview storage and reads

wesm:
Docbank can now retain a canonical visual preview result for an exact content version and let embedded applications inspect or stream it. This gives photo and video applications one shared derivative identity and lifecycle instead of forcing each application to build its own preview catalog.

A ready result records the exact output bytes, media type, and dimensions. Unsupported formats and deterministic decoding failures remain queryable outcomes without pretending that preview content exists. The complete recipe is fingerprinted so decoder, scaling, color, orientation, frame-selection, or encoder changes create a new immutable result.

Preview metadata and bytes participate in backup and restore, integrity validation, derivative reporting, and garbage collection. Deleting the source content version releases its preview catalog, while content-addressed storage still deduplicates identical preview bytes across versions.

The embedded API now exposes preview status and verified preview reads. This change establishes the storage contract only; automatic image, camera RAW, and video producers will follow separately.

### DIFF
--- a/cmd/docbank/versions.go
+++ b/cmd/docbank/versions.go
@@ -201,7 +201,7 @@ var versionsPruneCmd = &cobra.Command{
 
 func writeVersionPruneReport(cmd *cobra.Command, report api.VersionPruneReport) {
 	_, _ = fmt.Fprintf(cmd.OutOrStdout(),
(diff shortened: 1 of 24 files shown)


wesm:
invalid premise

PR #253, opened 2026-09-02
Read creation time from MP4 files

wesm:
Docbank now records the creation time declared by an MP4 movie header as canonical source metadata. Embedded applications can use the original video chronology without maintaining another container parser.

The public media detector validates and exposes the timestamp, and source-metadata extraction publishes it as the standard `created` fact. Missing, sentinel, and unrepresentable values stay absent.


wesm:
false positive

PR #254, opened 2026-09-02
Fix and consolidate the post-0.14.0 processing foundation

wesm:
Fixes the problems a full review found in everything merged after v0.14.0, and merges the copies that the rest of the processing stack would otherwise keep repeating. None of this has shipped, so schema and contract changes are direct, with no migrations.

**What is different for users**

- A vault no longer refuses to open after a harmless edit to `schema.sql`. The schema is declared once instead of fingerprinted by its own text.
- Search indexing no longer grows with the square of the corpus. Text is indexed once per build and old generations are collected when the head moves.
- `docbank gc --run` no longer fails with a 500 after an interrupted run.
- `versions prune` reports the same reachability that gc uses, so it no longer calls pinned bytes releasable.
- Opening a store no longer rescans every cached text.
- `docbank jobs` shows `process:renditions` only when a provider is bound.

**What is different for the code**

- One provider request layer under `document/internal/providerutil` for all six adapters, with one status mapping and `Retry-After` support. Constructors are unchanged.
- One canonical JSON codec, one daemon backfill loop, one publish transaction, one blob reachability rule.
- Metadata values now round-trip, unit limits apply while rendering, permanent seal errors stop retrying, and the extractor fingerprint is pinned by a test.

Review focus: the lexical index redesign in `internal/store/search.go` and `schema.sql`, and the shared status table in `providerutil/status.go`, which changed three adapter test expectations.

🤖 Generated with Claude Code (claude-fable-5-1)
Co-authored-by: Claude Fable 5.1 <noreply@anthropic.com>


wesm:
cc @salmonumbrella

PR #258, opened 2026-09-03
Produce canonical previews for GIF originals

wesm:
Docbank can now produce its canonical JPEG preview from still and animated GIF originals. Embedded applications get the same exact-version identity, verified reads, and storage lifecycle already available for JPEG and PNG sources.

Animated GIFs deliberately use their primary frame. A partial primary frame is placed on the logical canvas, and transparency is composited onto white before scaling. Malformed GIFs become durable failures, while source-read failures remain retryable. The changed processor identity causes earlier unsupported GIF results to be produced again when requested.


wesm:
invalid

PR #259, opened 2026-09-03
Produce canonical previews for WebP images

wesm:
Docbank can now turn still WebP originals into the same retained JPEG preview it provides for JPEG, PNG, and GIF. Embedded applications can display these images without decoding authoritative originals themselves.

WebP previews honor EXIF orientation, flatten transparency onto white, and enforce the existing pixel and metadata limits. Files with embedded ICC color profiles remain unsupported until Docbank has managed color conversion. Animated WebP is also reported as unsupported instead of producing an incorrect still.

The processor identity changes with WebP support, so earlier unsupported results are produced again when requested.


wesm:
invalid

Issue #175, opened 2026-08-22
Authorize bounded non-PDF formats in document/mistral


salmonumbrella:
## Problem

Docbank's current public `document/mistral` contract can authorize only PDF because its page limit is the only provider-unit cap the client can enforce before upload. That keeps Msgvault's shared Docbank-backed extraction path PDF-only after kenn-io/msgvault#635.

kenn-io/msgvault#608 targets every safely supported non-visual document family. kenn-io/msgvault#664 adds document vectors and semantic/hybrid search for chunks that already exist; it does not broaden Mistral extraction authority.

## Existing evidence

A synthetic authenticated probe against the pinned EU `mistral-ocr-4-0` endpoint exercised 26 candidate formats.

Provider-accepted (18): PDF, DOC, ODT, RTF, PPT, XLSX, CSV, EPUB, TXT, Markdown, RST, JSON, JSONL, YAML, Go, Python, JavaScript, and MSG.

Provider-rejected (8): DOCX, PPTX, XLS, ODS, Numbers, LaTeX, XML, and EML.

That is discovery evidence, not upload authority. A provider accepting a file does not prove Docbank can enforce the declared page, slide, sheet, record, or other provider-unit bound before bytes leave the machine.

## Ask

Extend the storage-neutral Docbank packages so each non-PDF format is authorized only when its limits can be enforced locally before upload:

- detect the real container and family rather than trusting caller metadata;
- enforce bytes plus a family-appropriate unit bound before staging or upload;
- record per-format probe evidence and policy identity in the capability manifest;
- keep formats that cannot be bounded explicitly blocked with a stable reason;
- ship deterministic synthetic fixtures and public contract tests for every enabled family;
- preserve the existing private staging, redirect refusal, bounded response, retry, and sanitized-error guarantees;
- remain stateless and inline-only—never fall back to Mistral Files or Libraries.

The consumer contract should let Msgvault pin Docbank and enable an authorized format without duplicating Docbank's detector, bounds, probe, or transport.

## Acceptance

- Every provider-accepted format that can be bounded safely has explicit runtime authority and family-specific tests.
- Every remaining format fails closed before upload with a stable operator-visible reason.
- Capability and policy fingerprints change when any format, bound, endpoint, model, or request policy changes.
- No Docbank vault, daemon, database, queue, vector store, or search dependency enters the public document-processing packages.

Msgvault continues to own attachment/message provenance, consent records, orchestration, persistence, vectors, and search serving.

[...]

PR #261, opened 2026-09-03
feat(rendition): carry local unit counts through Mistral OCR

rodboev:
The Mistral OCR path accepted output without checking that the provider returned the number of text units the local inspection counted, and the Mistral seam was not ready to claim authority for non-PDF formats. It now carries the verified local unit count through rendition snapshots and compares it with the returned units before output is accepted, leaving the seam ready for non-PDF authority once provider evidence proves the unit semantics.

Unrecognized, malformed, over-limit, and unproved formats continue to fail closed; the existing PDF authority, exact-byte checks, and consent boundary remain unchanged. Callers that need a non-PDF input today must convert it to PDF, while provider evidence is recorded from synthetic inputs or a credentialed probe rather than inferred from hand-written responses.

The change stays inside the Mistral provider adapter and bounded format inspection; other providers, storage, daemon, API, CLI, and UI surfaces remain unchanged.

Refs #175


wesm:
looking

wesm:
I checked the findings and made the small cleanups. The title now says that this PR carries local unit counts through Mistral OCR. It does not enable any new non-PDF formats.

The extra non-PDF mismatch check was redundant: the OCR client rejects that response earlier. I restored the PDF-only check and combined the duplicated source-count validation while preserving the PDF error messages. The DOCX test now builds its ZIP once and checks both a matching response and rejection of three units against a local count of two.

I kept the temporary test registration inline and documented that this test must stay sequential. A one-use helper would not enforce that constraint. I also kept the registry test, with a note that it guards future registrations and currently has no registered counters to exercise.

Pushed in b8caef1a. The full Go suite passed with CGO enabled and disabled; the Mistral race test, lint, and prek checks also passed.

Issue #200, opened 2026-08-24
add: replacing an existing path creates 'name (2)' instead of a version; no time-window listing without a query


Joi:
Two more from running docbank as the system of record behind a mirrored vault + Google Drive (follow-up to #197).

## 1. `add` onto an existing path suffixes instead of versioning

Re-running `docbank add edited.pdf --dest /vault/x` where `/vault/x/edited.pdf` already exists with different bytes creates `/vault/x/edited (2).pdf` (the documented collision suffixing). For an import that is re-run nightly against a live tree, that silently splits history: the original node keeps the old bytes and a new node appears beside it. `docbank put` does the right thing, so my wrapper now diffs size/mtime against a state file and calls `put` for changed files, `add` only for new ones.

Suggestion: an `add --replace` (or `--reconcile`) mode: when the destination path resolves to an existing file node, treat the source as a new version of that node (same `If-Match` semantics as `put`, fail closed on a directory). Unchanged bytes would still be skipped by hash. That makes a plain re-run of `add` converge instead of drift.

## 2. No way to list "what changed since T" without a query term

`GET /api/v1/search` requires `q`; `modified_since`/`modified_before` are filters on top of it, so "every document modified this week" is not expressible (a wildcard `q` is not accepted either). I wanted this for a morning-brief line; the workaround is an ingest journal kept by the wrapper.

Suggestion: allow an empty `q` when a time window (or `tag_id`) is present, returning the same bounded, truncation-flagged page ordered by `modified_at` desc.

Both are quality-of-life; nothing incorrect. docbank 0.14.0, macOS 26.5 arm64.

salmonumbrella:
@Joi build it!

rodboev:
The queryless search and replace-mode ingest slices have landed in #262 and #274.

PR #262, opened 2026-09-03
feat(search): allow bounded queryless filtered search

rodboev:
Search can now list documents by a tag or modification-time window with no search term, returning a bounded, truncation-flagged page ordered newest first and marked as a filtered result. Before, every search required a query, so "everything modified this week" could not be expressed (#200).

Blank searches with no tag or time bound are still rejected, and media-type and subtree filters still narrow an anchored request rather than becoming an open vault listing. Filter-only pages may scan live nodes before applying the limit, so a narrow window or tag is recommended on large vaults. This is the second half of #200.

Refs #200


wesm:
Pushed the agreed fixes:

- Brought the branch up to date with main and advanced the daemon protocol to 64, so the new CLI replaces an incompatible daemon even when their application versions match.
- Made tag and time searches use indexes to select results. The time index also covers the name and ID tie-breakers.
- Resolved all paths for a filter page in one database statement, removing the two extra queries per result.
- Gave a missing-query error its own API code and preserved it through the client and CLI. Kept the early CLI check so invalid commands fail before starting the daemon.
- Moved filter-only search into its own method and corrected its error messages.
- Clarified that whitespace-only queries count as blank, results include directories, and a truncated page may be impossible to complete using time bounds when many nodes share one timestamp.

Both complete Go test suites passed, with and without CGO. Lint, the strict docs build, and pre-commit checks passed. A synthetic 10,000-node check confirmed index use in both SQLite drivers and exercised a 1,000-result page.

Issue #218, opened 2026-08-26
Writable provenance: let an agent record an origin fact after ingest


Joi:
Provenance facts are recorded at ingest only (`GET /nodes/{id}/provenance`; the capability page says 'record immutable provenance facts and later corrections', but the HTTP contract exposes no way to add one).

My case: a triage tool moves documents from a laptop's ~/Downloads and ~/Documents into a Google Drive hierarchy; the nightly reconcile then ingests them from the Drive mount, so the node's provenance says `original_path = <Drive mount path>` and nothing about where the file actually came from, when it was moved, or by what. I keep that in a JSONL journal and stamp the node with tags (`origin:downloads`, `triage:joimba-2026-08-26`), which works but is a weaker record than a provenance fact.

Suggestion: `POST /api/v1/nodes/{id}/provenance` accepting a fact with the same shape as ingest facts — `source_kind` (e.g. `agent`), `source_description`, `original_path`, `original_mtime`, optional `supersedes` — appended immutably with `If-Match`, and auditable like other mutations. That would let filing agents leave a first-class trail without inventing tag conventions.

docbank 0.14.0. Related: #197, #200.

salmonumbrella:
@Joi go for it, feel free to add it in.

rodboev:
Writable provenance has landed in #273.

PR #273, opened 2026-09-04
feat(store): append provenance after ingest

rodboev:
After Docbank imports a file from a Google Drive mount, a filing agent can now add its original laptop location to the document's provenance through the HTTP or embedded Go API. Previously, provenance recorded only what was known at ingest, leaving agents to keep later evidence in tags or a separate journal.

Each append preserves document content and advances the node revision once. HTTP callers supply the current revision in `If-Match`; embedded callers can omit that check. When auditing is active, the fact and its audit records commit together. Source references remain evidence that Docbank stores without opening.

Corrections append a successor to an active caller-supplied fact on the same document and preserve the earlier fact in history. CLI and watched-folder ingest facts stay active so subsequent imports recognize the existing document. Correcting provenance supplied through embedded `Create` makes the original request return `ErrContentConflict` unless another active fact still matches.

Refs #218

### DIFF
--- a/docs/agents/integration.md
+++ b/docs/agents/integration.md
@@ -539,6 +539,22 @@ Provenance is evidence, not ownership of the external source. Reading it does
 not open or modify that source, and it does not make a content version a
 retention root.
 
+To record an origin learned after ingest, send the node revision returned by a
+prior read:
+
+```bash
+curl --fail-with-body -X POST \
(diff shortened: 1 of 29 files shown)


rodboev:
Rebuilding this one to save a couple hundred lines. Will re-derive request validation, and drop receipt recomputation and the overflow guard.

wesm:
looking

PR #274, opened 2026-09-04
feat(ingest): add a replace mode to local add

rodboev:
`docbank add --replace` now makes a file whose destination path already exists a new version of the existing document, instead of dropping a second copy beside it as `notes (2).pdf`. That suffixing quietly split a document's history when an import ran nightly against a changing tree; with `--replace`, unchanged bytes are still skipped by hash and the change is fenced by the node revision so a concurrent edit fails cleanly.

Suffixing stays the default, since two files sharing a name in a bulk import are usually different documents. HTTP integrations get the same option as `"replace": true`. This is the first half of #200; listing documents by time window with no query is separate work.

Refs #200

### DIFF
--- a/cmd/docbank/add.go
+++ b/cmd/docbank/add.go
@@ -3,6 +3,7 @@ package main
 import (
 	"encoding/json/jsontext"
 	"encoding/json/v2"
+	"errors"
 	"fmt"
 	"io"
 	"path/filepath"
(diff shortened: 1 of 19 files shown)


rodboev:
Rebuilding this one to match the opt-in flag design, despite the clean result from roborev. The issue has it right and the shape here should match that. Will mark it ready for review once it does.

wesm:
looking

Issue #240, opened 2026-08-30
Acquire, transcribe, and index message media from bytes and shared recording URLs


salmonumbrella:
## The situation

Voice notes and shared recordings are message content, not decorative attachments. Msgvault can already identify or retain media from several sources, but each consumer should not have to build its own downloader, transcript store, Docling adapter, embedding jobs, and retrieval index.

The ownership boundary should match #176: the source application knows which message the media belongs to and whether that message is still visible; Docbank handles the media itself. For a local voice attachment, Docbank receives the exact bytes. For a Loom or Cap link, Docbank receives a protected provider reference and, when the content is publicly or explicitly authorized, resolves and downloads it. Docbank then extracts or transcribes it, retains normalized evidence, indexes it, and returns one stable identity plus coverage state to the caller.

## What works today, and why it is not enough

- The #176 stack gives Docbank sealed exact-byte processing, durable jobs, Docling renditions, normalized evidence, embeddings, source-fenced retrieval, and embedded/HTTP application boundaries.
- The pending Docling adapter accepts bounded supplied files through fixed routes. Docling also supports audio/video transcription, but Docbank does not yet expose media transcription as a first-class processing profile.
- The pending Gemini adapter can embed WAV, MP3, MP4, and MOV directly. That is useful for multimodal similarity, but it does not produce the plaintext transcript needed for exact phrase and lexical search.
- Docbank has no provider-neutral remote-media source, no Loom/Cap resolver, and no durable distinction between metadata-only, access-required, downloadable, unsupported, and failed acquisition.
- Loom has no open management API. Its oEmbed surface provides metadata, while downloads depend on the recording and viewer's permissions.
- Cap's Developer API covers videos owned by the calling app, not every received share link. Self-hosted Cap can live on an operator-defined origin and needs an explicitly scoped trust and credential boundary.

Without this layer, Msgvault would have to duplicate processing and provider acquisition logic that belongs in Docbank, and every later consumer would have to do the same again.

## Ask

Add one durable media-ingestion lane to Docbank for supplied attachment bytes and authorized shared-recording URLs. It should end in the existing rendition, evidence, embedding, and source-fenced retrieval system rather than creating a parallel transcript store.

### 1. Accept two source shapes

- **Supplied media:** accept an exact bounded byte stream plus filename, declared MIME type, source digest, caller-owned source identity, and visibility fence. Seal and inspect the bytes before choosing a processing profile. The caller remains the authority for the original message attachment.
[...]

PR #304, opened 2026-09-09
feat(document): index supplied audio with its provider transcript

rodboev:
Docbank can pass a caller-held transcript through the rendition contract for supplied WAV and MP3 bytes. The provider keys transcript lookup by the authorized audio digest and retains the provider transcript artifact.

The existing transcript builder now exposes its source-evidence step. The builder and provider call that shared implementation, so validation, artifact bytes, and checksums stay aligned. Existing media inspection, upload authorization, byte verification, and error classification remain the owners of those checks.

This package supplies transcript text. Remote recording acquisition, Loom and Cap resolution, speech recognition, and public ingestion and retrieval remain separate work under #240. The provider includes the caller's source binding in its descriptor identity, so each source gets a distinct runtime identity. The implementation builds on [PR #295](https://github.com/kenn-io/docbank/pull/295).

Refs #240


mariusvniekerk:
Seems like a LOT of code?

mariusvniekerk:
This also seems to change a hell of a lot of the documentation. Is that intentional?

rodboev:
Hmm should be 6 files, +714/-17, docs are only 8 lines. Need to switch the base branch, sec.

wesm:
looking

PR #306, opened 2026-09-09
feat(search): save named queries and highlight sets

### LINKED ISSUE #305: Save named queries and highlight definitions in the vault
## The situation

A client that reuses a search or a set of literal highlights needs somewhere durable to keep the definition.

## What works today, and why it is not enough

Clients can keep their own copies outside the vault, but those copies are not included in vault backups and have no shared revision check when edited.

## Ask

Add named query and highlight definitions with stable IDs, canonical payload identity, revision-checked changes, and backup/restore support. Keep execution and rendering separate, and preserve the existing audit mutation boundary.

salmonumbrella:
## What changed

- Save named queries and ordered literal highlight sets through the API. Definitions have stable IDs, canonical fingerprints, and revisions.
- Require the current revision when editing or deleting. Canonical no-op edits keep the existing revision, and saved definitions survive backup and restore.
- Keep saved definitions readable after audit is enabled. Changes return an explicit error because audited history does not support this mutation yet.

## Why

Clients currently have to keep reusable queries and highlight definitions outside the vault. Saving them with the document catalog gives clients one durable definition to share and protects edits from stale writes.

## Usage

Create a definition with `POST /api/v1/saved-queries`:

```json
{
  "name": "Invoices",
  "kind": "query",
  "payload": {"v": 1, "text": "invoice", "filters": {}}
}
```

List definitions with `GET /api/v1/saved-queries`. Read, edit, or delete one at `/api/v1/saved-queries/{id}`; send its `ETag` as `If-Match` for edits and deletes. Use `kind: "highlight_set"` with a payload of `{"v":1,"terms":[{"text":"invoice","color":"#ffff00"}]}` to save literal highlights.

This stores definitions only. Query execution and highlight rendering are separate work.

Closes #305

Current web search interface, captured from this branch with synthetic documents. Saved definitions are available through the API.

Web search with synthetic documents


wesm:
reviewing now

PR #309, opened 2026-09-10
feat(email): decode MIME into verified metadata

salmonumbrella:
Imported `.eml` files keep their original bytes while Docbank records verified MIME structure, decoded headers, and searchable text from the selected outer message body. API clients can read or request metadata at `/api/v1/versions/{version_id}/email` and download individual parts through the resulting generation.

Go's standard library handles addresses, encoded headers, and multipart parsing. Source metadata uses the same decoder. Generations record the source identity, Go parser version, and fixed resource limits so repeated decoding can reuse verified results. Exact raw headers and decoded part bytes remain available with size and digest evidence.

Only the selected outer body enters search. Headers, attachments, and nested forwarded messages stay outside that index. Email routes require the master API key and exclude browser sessions. Email derivatives participate in backup, restore, purge, and garbage collection.

Malformed part headers and multipart delimiter or preamble lines longer than 4 KiB produce a partial inventory with an explicit diagnostic. Original bytes and already inventoried parts remain readable. Attachment indexing and a browser email viewer remain outside this change.


wesm:
working

wesm:
looking again

wesm:
still working on this

PR #310, opened 2026-09-10
feat(web): manage saved queries and highlight sets

salmonumbrella:
Manage saved queries and reusable highlight sets from the web app's bookmark button. Create, edit, rename, and deliberately delete definitions using the saved-definition API from #306.

**Keep query draft** retains the complete expression, filters, mode, and sort in the tab and URL fragment. Saving or deleting another definition leaves that draft alone. Query URLs contain search text and filters, so treat them as private. Locking or a rejected session clears the draft; save a named definition before opening a new session if you need it later.

Edits and deletions use the inspected revision. Stale writes remain visible without automatic retries. If a successful mutation returns a receipt the browser cannot verify, the catalog reloads and the editor requires a fresh selection before another save. Server-accepted Unicode names remain accessible in the catalog.

Saved-query execution and highlight rendering remain unavailable. Keeping or saving a definition does not change live results. Highlight sets contain literal terms and lowercase hex colors. No schema changes or new query executor.

Saved-query editor in the actual web app, using synthetic proposals.

Reusable literal highlight terms in the actual web app, using synthetic data.

### DIFF
--- a/docs/usage/searching.md
+++ b/docs/usage/searching.md
@@ -133,6 +133,9 @@ Save a named search definition when several clients need to reuse it. The
 HTTP API stores the definition with the vault's metadata. It does not run the
 search. Saved definitions survive backup and restore.
 
+The [web management screen](web.md#saved-queries-and-highlights) also manages
+saved definitions.
+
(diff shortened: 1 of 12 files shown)


wesm:
looking

PR #311, opened 2026-09-11
feat(web): browse and label import collections

salmonumbrella:
## What changed

- Browse import collections and their current documents from the web app. Each collection shows its label, import time, live document count, and logical bytes.
- Rename or clear collection labels without changing documents. A conflicting edit keeps your draft and asks you to reload instead of overwriting another client's change.
- Open a member by its inspected document identity. Collection lists and member lists show the first 100 entries and explicitly report when more are available.

## Why

Import collections are available through the API, but browser users need a way to find and label a group of imported documents without switching tools.

## Usage

Choose **Import collections** in the top bar, select a collection, and open a document or edit its label. Label writes are unavailable after permanent audit authority is enabled.

This browses collection membership directly; it does not filter text search. Processing quality and coverage remain explicitly unavailable. Use the paginated HTTP API for entries beyond the browser's limit.

Built on #308. No schema changes.

Import collection members in the actual web app, using synthetic documents.

A conflicting label edit preserves the draft in the actual web app.

### DIFF
--- a/docs/usage/web.md
+++ b/docs/usage/web.md
@@ -473,6 +473,28 @@ still readable. Run `docbank backup verify` to independently prove repository
 integrity, and periodically restore into a separate vault to rehearse the
 complete recovery path.
 
+## Browse import collections
+
+Choose **Import collections** in the top bar to browse the vault's import groups.
+Each card shows its label or source description, ingest time, current live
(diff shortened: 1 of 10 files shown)


wesm:
rejected

PR #312, opened 2026-09-11
feat(search): compile field and Boolean expressions

salmonumbrella:
## What changed

- Preview field-aware search expressions with phrases, Boolean groups, prefixes, and proximity matching. The response preserves the complete query and separate filters, with the identities and revisions of referenced tags, collections, and saved queries.
- Keep saved-query filters inside their own expression scope. Missing references, cycles, unsupported combinations, and oversized expressions produce explicit errors instead of silently broadening a query.
- Compile parameterized predicates over the existing name and text indexes, preserving current document versions and the selected rendition generation. Both SQLite drivers support the same predicates, including nanosecond time bounds.

## Why

Saved definitions can retain richer query intent than the existing simple-search route can execute. Clients need to validate that intent and explain errors before using it, without losing field constraints or treating a missing excluded tag as a successful query.

## Usage

Send a QueryV1 object to authenticated `POST /api/v1/queries/parse`, for example:

```json
{
  "syntax": "advanced",
  "text": "name:(budget OR forecast) AND NOT extension:tmp",
  "filters": {"size_min": 100}
}
```

The response contains `query`, `query_fingerprint`, and `dependencies`. Expression errors return `422 invalid_query` with a UTF-8 byte span; database failures remain server errors.

This is validation and compilation, not a new search executor: it returns no result rows or SQL, and the CLI and existing `/search` route keep their current behavior. Snapshot execution is separate work. Semantic/hybrid mode, relevance ordering, duplicate constraints, and text-coverage constraints are explicitly unsupported here.

Bare dotfiles such as `.pdf` have no extension in both the extension filter and media-family fallback. A declared MIME type still takes precedence; `.report.pdf` retains its extension.

Built on the saved-definition and collection APIs in #306 and #308. No schema changes or second lexical index.

Existing search interface, captured with the repository's synthetic documentation vault. Query preview is available over HTTP; this change adds no controls to this screen.

Docbank search results with synthetic documents

### DIFF
--- a/cmd/docbank/daemon_test.go
+++ b/cmd/docbank/daemon_test.go
@@ -7,6 +7,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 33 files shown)


mariusvniekerk:
Doing a bit of a refinement pass on the impl here

wesm:
## Review of PR #312

**Findings:** 0 P1, 0 P2, 3 P3, 4 P4. All three P3 items are fixed in 03413181. The P4 items are noted only.

### Fixed (P3)

- **Windows amd64 CI**: three daemon tests in `cmd/docbank` failed with "daemon did not shut down" and health checks that never passed within ten seconds. The package took over five minutes on that runner. Two of the tests also returned without joining the daemon, so `TempDir` removal raced a live SQLite file and produced a second, misleading error. The tests now share one start/stop helper that waits up to a minute for startup, thirty seconds for shutdown, and always joins the daemon before the test ends.
- **NEAR distance parsing**: `NEAR/+5` was accepted because `strconv.Atoi` accepts a sign, while `size_max:+1` was rejected. Both now require plain digits.
- **Empty error position**: a preview whose query failed structural validation (not an expression error) returned `position: {offset: 0, end: 0}`. The handler now omits `position` when there is no byte span to report.

### Noted (P4, not changed)

- `docbank_query_timestamp_v1(n.modified_at)` re-parses every row even though the store writes `modified_at` as fixed-width UTC nanoseconds, which already sorts lexicographically. Comparing the column directly against the normalized key would avoid a per-row function call and keep any future index usable. Left as-is since it is preview-only today and the function guards against older rows.
- `Store.CompileQuery` (method) and `store.CompileQuery` (package function) share a name with different signatures. Legal, but easy to misread.
- The stabilization changes to `document/mistral/rendition_test.go` and `internal/extract/worker_test.go` are unrelated to the feature. Fine to keep here since they unblock CI, but they widen the diff.
- `docs/architecture/http-api.md` reorders the frontmatter keys without a content change.

### CI status

- **Linux tests** fail before any test runs: Docker Hub denies the `minio/minio` pull. The PR workflow uses `ci.yml@main`, so this branch cannot fix it. Open PR #325 switches the image to Quay; once it merges, a rerun here should pass that job.
- **Windows amd64** is addressed above. Other recent PRs fail different Windows tests on slow runners, so some flakiness may remain outside this PR's files.

### Verification

Local: `go build`, `go vet`, `golangci-lint`, and `go test -tags fts5` pass for `cmd/docbank`, `internal/query`, `internal/store`, `internal/api`, and `sqlite`. Windows behavior is unverified locally.

PR #316, opened 2026-09-11
Add document keyboard shortcuts

salmonumbrella:
Browse the loaded document list and toggle tags without moving between the mouse and keyboard.

Press `?` for shortcut help. `/` focuses search, `j` and `k` move between loaded rows, Space toggles file selection, Enter opens the inspected folder, and Escape clears selection. Assign digits `1`–`9` to tags in the help dialog; bindings stay separate for each vault and must be set again after the daemon restarts or browser data is cleared.

Navigation stops at the loaded-page boundary. Tag shortcuts affect only the inspected file, regardless of checked rows, and use its displayed revision. A conflict requires a refresh; it never silently retries against newer state. Shortcuts pause during text entry and dialogs.

Keyboard shortcut help with a synthetic document and tag binding

### DIFF
--- a/cmd/docbank/cli_test.go
+++ b/cmd/docbank/cli_test.go
@@ -80,19 +80,9 @@ func setupVaultHome(t *testing.T) string {
 // production; no test-only transport exists.
 func startTestDaemon(t *testing.T, dir string) {
 	t.Helper()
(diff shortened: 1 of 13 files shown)


wesm:
false positive

PR #317, opened 2026-09-11
Apply tags atomically to selected documents

salmonumbrella:
Add or remove a tag from selected documents in one operation. Select files on the displayed page and choose **Edit tags**; the dialog shows whether all, some, or none already have the chosen tag.

Each operation covers an explicit set of at most 1,000 documents. Every displayed revision must still match: either the whole set succeeds or nothing changes. Folders and undisplayed search results are never swept into the operation. Stale or deleted selections require a refresh; audit restrictions and operation-ID collisions retain their own errors.

If the response is lost, **Retry same operation** reuses the original request and returns its retained receipt without applying the change twice. Receipts use a fixed v1 format and survive restart, metadata export/import, and backup/restore. The database reads the selected targets together and applies bulk writes. Permanent audit records remain per document, with separate audit operation IDs and a shared batch timestamp.

Selected-document tagging with a mixed tag assignment on synthetic documents


wesm:
looking

PR #318, opened 2026-09-11
Edit and validate complete search queries

salmonumbrella:
Edit complete search queries without losing their filters, mode, or sort order. Start in **Edit query**, open a saved query, or create a draft for an import collection. Keep the full draft in the tab and URL, or save it as a named definition. Malformed drafts must be corrected or explicitly discarded before closing the editor.

The server validates each draft as you type. Syntax help and highlighted errors help you correct expressions, and newer edits supersede older validation responses. **Run query** remains unavailable; validation does not execute a search or drop constraints to fit ordinary live search.

Complete query validation with a synthetic collection filter

An unknown tag exclusion stays an error and is highlighted in the editor

### DIFF
--- a/docs/usage/searching.md
+++ b/docs/usage/searching.md
@@ -216,6 +216,16 @@ in a vault that needs editable saved definitions.
 
 ## Preview a field-aware query
 
+In the web application, choose **Edit query** for expression editing, structured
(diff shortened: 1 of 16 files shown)


wesm:
rebasing

PR #320, opened 2026-09-11
Inspect collection processing coverage and quality

salmonumbrella:
See which documents in an import collection are searchable and inspect its file types, sizes, dates, duplicates, and empty files without exporting it. Processing details load when you inspect quality, so you can browse collections even when processing data is incomplete.

Coverage follows each document's current version and the selected processing profile. It distinguishes complete, partial, failed, unprocessed, and no-text output, shows missing configuration separately, and retains searchable output after a failed retry.

Select extensions, media types, or media families to open a query draft scoped to that collection. Suggestions leave live results unchanged; **Run query** remains unavailable. Large or interrupted quality reads return an error instead of partial totals.

Collection quality for a synthetic import, with unavailable processing shown separately

A quality suggestion opens a collection-scoped query draft and passes validation


wesm:
rebasing

PR #321, opened 2026-09-11
Open consistent, bounded query snapshots

salmonumbrella:
Page through query results while their rows, totals, and facets stay fixed as documents change. The snapshot API supports forward and backward paging, duplicate filters, and processing coverage.

Snapshots belong to the requesting session, hold up to 250,000 rows, and expire after 15 minutes idle or 30 minutes total. Restarting the daemon clears them. An unavailable snapshot returns an explicit error; paging never silently reruns a live search.

Saved-query runs require the inspected definition revision and record totals and changes from the previous run. Those records survive backup and restore; cached rows remain temporary.

Semantic and hybrid queries and relevance ordering remain unsupported. The browser query editor still validates drafts without displaying snapshot pages.


wesm:
rebasing

PR #322, opened 2026-09-11
feat(web): browse and act on query snapshots

salmonumbrella:
Run lexical queries in the web app, browse frozen results with exact counts and facets, and tag selected rows or the whole query. Failed requests keep accepted results available. Later tag changes appear alongside the frozen rows without changing their membership. Leaving or replacing a snapshot cancels unfinished action preparation.

Whole-query tagging requires a saved recovery file and confirmation of the exact targets. Interrupted actions can resume with their original requests. Imported recovery files require the daemon to confirm each batch before it counts as complete. Revision conflicts and missing tags or documents keep the action stopped across browser tabs. Deleted tags and unreadable journals can be abandoned explicitly. Recovery-file checksums detect inconsistent data; they do not authenticate the creator, so imported targets must be reviewed before confirmation.

Verified with the real daemon and 1,001 synthetic documents, including selection-then-whole-query tagging, lost responses, restart recovery, deleted tags, and corrupt journals.

Builds on #321 (merged).

Frozen query workspace with synthetic documents

### DIFF
--- a/docs/architecture/http-api.md
+++ b/docs/architecture/http-api.md
@@ -61,6 +61,7 @@ Endpoints are filesystem-shaped, under `/api/v1`:
 | `PATCH /nodes/{id}` | move and/or rename, including resolving an absolute `dest_path` transactionally | Implemented |
 | `POST /path/move` · `POST /path/trash` | move / trash by virtual path, resolved and mutated in one store transaction | Implemented |
 | `POST /batch/move` | validate and apply up to 1,000 moves as one final-state transaction | Implemented |
+| `POST /batch/tags` · `POST /batch/tags/preview` | atomically assign/remove one tag on a revision-fenced selected set, or inspect its exact membership | Implemented |
 | `POST /nodes/{id}/trash` · `POST /nodes/{id}/restore` | soft delete / recover | Implemented |
 | `GET /trash` · `POST /trash/empty` `{run, older_than}` | list (optionally paginated) / report or hard-delete trash roots | Implemented |
 | `POST /gc` `{run}` · `POST /verify` | reclaim unreachable blobs / validate metadata and re-hash all blobs | Implemented |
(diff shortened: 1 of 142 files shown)


wesm:
rebasing

wesm:
Snapshot browsing now stays usable after a failed page request, and clicking a selected month clears that filter. Discarding a draft and reopening the editor restores the accepted snapshot query. Live document controls stay hidden in snapshot mode.

Selection tagging now carries your chosen tag and add/remove action into confirmation. Recovery only treats an HTTP 401 as an expired session, so ordinary errors no longer log you out.

The frontend checks and regression tests pass. The journal and recovery-file fixes are still in progress.

wesm:
Recovery can now be opened and abandoned after its tag is deleted. The action menu can also abandon an unreadable journal without clearing another vault’s work.

Whole-query tagging now uses your confirmed selection changes. Revision advances must connect back to the frozen document, so an unrelated newer document version still stops the action.

Recovery files now show the exact target identities before confirmation. Their checksums detect inconsistent data; they do not authenticate the creator. File-supplied receipts no longer advance snapshot revisions.

Verified with the real daemon and 1,001 synthetic documents, including selection-then-whole-query tagging, deleted tags, corrupt journals, lost responses, and restart recovery. Frontend checks, full Go tests in both SQLite modes, docs, lint, and pre-commit checks passed. Receipt-processing cleanup follows separately.

wesm:
Whole-query actions now validate and store one batch receipt at a time. They previously reloaded and re-encoded the entire plan after each batch, which made larger actions do increasing amounts of unnecessary work.

I also removed the unused saved-snapshot path, duplicate query hashing, and circular imports. The fresh vault check stays: recovery must verify which vault the current session can change.

All 497 frontend tests, the real-daemon browser test, full Go tests in both SQLite modes, lint, docs, and pre-commit checks passed.

wesm:
A whole-query action now stays stopped after a revision conflict, even when another tab receives a late success or failure. Valid receipts still get saved, but neither response can make the action resumable again.

Added two browser tests with a real daemon and shared journal. They reproduce both races and confirm that the last batch is never sent. Both tests failed before the fix and pass now.

Frontend checks (497 tests), all three snapshot browser tests, full Go tests in both SQLite modes, release-script tests, lint, and pre-commit checks passed.

wesm:
Whole-query preparation now stops when you cancel, leave or rerun the snapshot, or lock the session. Late responses and file reads from a dismissed dialog cannot reopen recovery. Cancellation also aborts journal writes that have not committed; already-saved journals remain available for explicit recovery or abandonment.

If a tag or target document is no longer available, the action stops instead of offering repeated retries. The recovery dialog explains how to abandon it and start again with a fresh selection.

Regression checks cover delayed responses, dismissed file imports, cancellation during a real IndexedDB transaction, and deleting a tag after recovery opens. The browser checks use the real daemon with synthetic documents.

wesm:
Tagging is now disabled while a snapshot loads. Rerunning or leaving closes the old selection and recovery dialogs, and their late responses cannot update the replacement. Closing recovery stops further batches; receipts for whole-query requests already sent remain saved.

Receipts now keep the newest observation for each tag. An older Review receipt arriving after a newer Archive receipt preserves both. Later whole-query actions advance document revisions only when the receipts account for every intervening change.

Explicit recovery in another tab still works. Its results belong to the snapshot displayed when recovery opens; replacing that snapshot retires the recovery view.

wesm:
File-supplied receipts no longer mark newly imported actions complete. After you confirm the targets, the app resends the original requests with their original operation IDs. Only receipts returned by the daemon count as completed work. Requests that already succeeded recover their saved receipts without applying the changes twice.

Both recovery file pickers now reject files larger than 128 MiB before reading them into memory.

The real-daemon browser check imports fabricated completion receipts for 1,001 synthetic documents and verifies that both batches still run. It also imports genuine completed receipts after a daemon restart and verifies that replay leaves document revisions unchanged.

wesm:
You can now resume an interrupted whole-query action from a recovery window that was already open. The retry keeps the original request and operation ID, so the daemon returns any saved receipt without applying the change twice. Paused, stale, and completed actions remain stopped.

The new browser regression closes the sending tab after the daemon commits, then completes recovery from the other tab. It checks the original request, receipt, and document revision. All four real-daemon browser tests, 512 frontend tests, both Go SQLite suites, lint, and pre-commit checks passed.

Fixed in 2e96bca6cfc25f7c4baf86844cd65e8add209ebb.

PR #323, opened 2026-09-11
feat(web): inspect verified document versions

salmonumbrella:
Inspect and download the exact document version selected in live results or a frozen query. Frozen previews keep their original bytes after replacement, with current metadata shown separately. Live results require a refresh before document actions if the selected content has changed or current metadata cannot be verified.

Text and PNG/JPEG previews verify the version, size, media type, and SHA-256 before display. Other files remain available through verified download. Switching documents or locking the session cancels previews and clears image resources.

Live tags load completely and refresh after snapshot tagging. Metadata-only changes update live details and action revisions. Open audit history refreshes when the inspected document changes.

Builds on merged #322.

Checked with the real daemon and synthetic documents, including replacement, cancellation, tagging, renaming, and download.

Verified original-version preview with separate frozen facts and live observations


wesm:
rebasing

PR #324, opened 2026-09-11
feat(web): read and navigate verified text

salmonumbrella:
Read extracted text, find terms, and move between documents without leaving a frozen query. Query highlights and saved highlight sets help locate relevant passages without changing the query. Choose 50, 100, or 250 documents per page before running a query.

Text stays tied to the selected version and verified rendition. Missing or changed renditions are reported explicitly. Processing status stays visible when verified original text is shown as a fallback. Opening a duplicate keeps the original query position available for return.

Builds on merged #323.

Verified PDF text with query and find highlights at the last document in a frozen query


wesm:
rebasing

PR #326, opened 2026-09-11
feat: add verified exact-version page images

salmonumbrella:
Request verified PNG page images for an exact document version through the API and typed clients. Page coordinates stay the same when DPI changes. Callers request up to 16 pages; imports do not render automatically.

Images stay tied to the selected source, page, and rendering settings. A new operation can retry failed or canceled work. Cancellation and stale selections block late publication. Partial images survive backup and restore and remain readable without a renderer.

Optional rendering requires Linux amd64 and pinned local executables. PDF verification allows a one-pixel rounding difference and records the actual image dimensions. PNG sources need explicit physical density and retain their original pixels. Fonts and shared libraries remain operator-managed. See [configuration and limits](https://github.com/salmonumbrella/docbank/blob/bc96a5f4bd1f7027ba7b582d2a3b337dfb72fe27/docs/architecture/page-images.md).

Builds on merged #324.

### DIFF
--- a/cmd/docbank/daemon.go
+++ b/cmd/docbank/daemon.go
@@ -207,6 +207,10 @@ func runServe(ctx context.Context) (retErr error) {
 		}
 	}()
 	operationGate := api.NewOperationGate()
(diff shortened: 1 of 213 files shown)


wesm:
rebasing

PR #327, opened 2026-09-11
feat(web): navigate verified document pages

salmonumbrella:
Read PDF pages in the document inspector with page navigation and fit-width or numeric zoom. The viewer prefers 144-DPI PDF images and lets you render one page at a time. PNGs with retained page geometry use their native density; other PNGs keep the verified inline preview.

Images are verified against the selected document version before display. Refreshing an unchanged selection keeps its page and zoom. Retained pages remain readable without a renderer, and verified-original download stays available when a source cannot be previewed.

Builds on merged #326.

Verified rotated page with navigation, zoom, partial inventory and original download


wesm:
rebasing

PR #329, opened 2026-09-12
feat(web): preview and download verified exports

salmonumbrella:
Export selected documents, the current page, or a whole frozen query as a verified ZIP. Preview the exact versions, included files, missing optional content, size estimate, and expiry before starting. Later query edits and document replacements do not change the reviewed export.

Download the completed archive with your chosen filename, final size, and SHA-256. You can rename the download without repeating the preview. Closing the drawer leaves the server job running; reopening it in the same browser session reconnects to that job. Original files remain unchanged and are not redacted.

The real-browser check exported 1,001 frozen versions after changing a document and the live query. All 1,004 ZIP entries, checksums, and captured versions matched the preview.

Builds on merged #328.

Export drawer with synthetic documents

### DIFF
--- a/cmd/docbank/daemon.go
+++ b/cmd/docbank/daemon.go
@@ -207,6 +207,10 @@ func runServe(ctx context.Context) (retErr error) {
 		}
 	}()
 	operationGate := api.NewOperationGate()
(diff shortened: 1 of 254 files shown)


wesm:
rebasing

PR #330, opened 2026-09-12
feat(email): publish and index exact attachment documents

salmonumbrella:
Applications can publish email attachments as ordinary documents, search their contents, and follow each attachment back to the exact source email version.

Publication is atomic and retryable. Repeated filenames and identical bytes keep separate occurrences unless the caller explicitly reuses a child version. Inline resources and explicitly attached single-part emails are included. Encrypted or unavailable parts remain visible in receipts without creating child files.

Receipts and relationships survive backup and restore. Retained references block permanent deletion; errors identify the blocking operation, and `docbank email-documents` lets CLI users inspect and release it while keeping child documents. Vault-wide derivative purges skip retained email inventories.

Go, HTTP, and typed client APIs expose publication and processing. HTTP operations require the master API key. Processing verifies the attachment bytes against a configured provider and requires explicit consent. Embedded calls wait for completion; HTTP requests enqueue work for the daemon. Relation status distinguishes searchable, partial, empty, pending, and failed output. Attachment text stays out of the source email body's search index.

Builds on #309 (merged).


wesm:
rebasing

PR #331, opened 2026-09-12
fix: let export verification finish and retire stale page jobs

salmonumbrella:
Large exports can finish verification before a download ticket is issued. Only valid POST download requests receive that timeout exemption; other methods and malformed download paths keep the usual timeout. Caller cancellation, caller deadlines, and the request-size limit still apply.

Regression coverage checks that changed or trashed documents release their old page-render job slots and retain published page results. Obsolete workers cannot retire replacement jobs, and database failures remain distinct from stale documents. This keeps the retirement behavior already merged upstream.

Builds on merged #329.


wesm:
rebasing

PR #335, opened 2026-09-12
feat(web): navigate exact email attachment versions

salmonumbrella:
Follow an email’s published attachments and parent messages from **Attachments**, including nested emails. Each link opens the exact attached version, even after its contents change. Return to the original result with its selection and focus intact.

Repeated filenames remain separate occurrences. Inventory completeness and processing status are shown separately, and missing targets stay visible. Navigation adds read access only; it does not publish attachments, start processing, or grant consent.

Builds on #334 and #331, both merged.

Verified historical attachment

Nested attachment and parent navigation


wesm:
rebasing

PR #336, opened 2026-09-12
Read archived email safely in the web inspector

salmonumbrella:
Read archived email in the web inspector with decoded or raw headers, HTML and plain-text alternatives, and clear warnings when MIME data is incomplete. Follow attachments at their exact versions and download the original EML.

The HTML reading copy blocks sender scripts, links, styles, and remote resources. Only verified local PNG and JPEG MIME parts display inline within fixed size limits; missing or ambiguous images keep visible placeholders. Quote folding leaves the original message and complete PDF source unchanged.

Builds on #335 (merged). Screenshots show synthetic messages on the real daemon.

HTML email with verified inline image and blocked resources

Long plain-text email

Partial MIME inventory with explicit warnings


wesm:
rebasing

PR #337, opened 2026-09-12
Keep mailbox imports watchable and retained trash cleanup progressing

salmonumbrella:
## What changed

- Keep mailbox watch streams open until completion or caller cancellation instead of ending them at the ordinary request timeout.
- Let trash cleanup delete unrelated items while preserving messages and attachments referenced by email publications. Report retained roots separately, including folders containing retained content.
- Enforce mailbox formats, states and size limits in Go across writes, reads and integrity checks, without baking evolving policy into the unreleased SQLite layout.

## Why

A long mailbox import should remain watchable, and one retained message should not prevent the rest of the trash from being emptied. Retention still protects the original identities, retry receipts and backup/restore records.

## Usage

Continue using `docbank mailbox watch <job-id>` and `docbank trash empty --run`. Trash reports now distinguish eligible, deleted and retained roots. Retained roots remain restorable; this change does not release their email-publication references.

Stacks on #335. [Repair-only diff](https://github.com/salmonumbrella/docbank/compare/78509ffd459a02f97daade069c1b374524f09f52...fix/mailbox-retention-boundaries).


wesm:
rebasing

PR #338, opened 2026-09-12
feat(export): bundle retained email PDFs and attachment occurrences

salmonumbrella:
## What changed

- Export retained email PDFs from a selection or a completed mailbox import, with optional original attachments and separately rendered nested-email PDFs.
- Preserve every message and attachment occurrence by default. Optional exact-content sharing reduces duplicate files while keeping every source receipt.
- Download one archive containing bounded, numbered ZIP volumes and a complete manifest. Preview shows unavailable outputs explicitly; partial export requires an explicit choice.
- Reuse verified retained PDFs only. Export does not render new PDFs, and unsupported attachment formats are never presented as rendered derivatives.

## Why

Exporting a mailbox should preserve its messages and attachment relationships without making users assemble individual downloads. The archive keeps exact source versions, recipes, checksums, and occurrence records together, including when identical outputs share a file.

A real browser run imported two synthetic messages with four attachment occurrences, rendered the parent and nested-email PDFs, and independently checked the downloaded ZIP contents and PDF text. The default export produced eight output volumes; explicit sharing produced four while preserving all occurrences.

## Usage

Open Export from a selection, or choose **Export completed collection** after a mailbox import. Select email body PDFs, find a retained recipe, choose attachment and duplicate policies, then preview and start the export. Missing requested outputs block export unless partial export is explicitly enabled.

Default occurrence-preserving export:

Email PDF export preserving all occurrences

Explicit duplicate-output sharing:

Email PDF export with explicit duplicate sharing

Stacks on #335. [Feature-only diff](https://github.com/salmonumbrella/docbank/compare/78509ffd459a02f97daade069c1b374524f09f52...feat/email-pdf-batch-export).


wesm:
rebasing

wesm:
Fixed all five findings:

- Messages with several attachment sets now offer an explicit choice, including empty sets. The export no longer leaves the drawer stuck on an unexplained conflict.
- **Start over** creates a fresh preparation after a failed source reservation. Ordinary retries still reuse their request identity.
- Loading another page of unavailable outputs keeps the plan's expiry timer and readiness. A page error stays beside the details instead of disabling a valid export.
- Body PDFs are selected from the same decoded version as the chosen attachment set. If no matching PDF exists, the requested missing-output policy applies.
- CSV rows for shared outputs now point to the file actually stored in the archive.

Focused regressions, both full Go suites (CGO and pure Go), all 766 frontend tests, the frontend build, strict docs build, lint, and pre-commit checks passed. A real-daemon browser check also exercised the attachment-set chooser with synthetic messages.

wesm:
Confirmed and fixed in bf537908.

The reviewed plan now becomes ready before the first unavailable-output details request finishes. If that request fails temporarily, the plan stays ready and its expiry timer keeps running. The drawer shows the details error with a retry button; retrying does not create a new plan. An expired-plan response still disables export start.

Regression tests cover the pending first request, temporary failure, retry, timer expiry, and HTTP 410. All 769 frontend tests, type checks, the production build, lint, and pre-commit checks passed.

Issue #333, opened 2026-09-12
Roadmap: document imports, review and exports


salmonumbrella:
The processing and retrieval foundation in #176 is the base for a broader document import, review and export workflow. This issue tracks the direction so each focused pull request can stay reviewable without losing the larger context.

## Outcomes

| Area | Goal |
|---|---|
| Imports | Mailboxes, Google Takeout Drive/Workspace, and broader file formats |
| Msgvault | Optional transfer across mail, chat, phone, calendar and meeting sources |
| Workspace | Tags, saved queries, selections, duplicate lookup and stable snapshots |
| Review | Exact document versions, text, pages, emails and conversations |
| Reports | Keyword-hit reports with clear counts |
| Exports | Verified bundles and PDFs |
| Access | The same workflows through web, CLI and MCP |

## Delivery sequence

1. Establish tag colors and groups (#299), page-local selection (#303), saved queries and highlight sets (#306), and import collections (#308).
2. Add saved-query and highlight management (#310) and import-collection browsing and labeling (#311).
3. Add field and Boolean query expressions (#312), duplicate-document lookup (#313), and a complete-query editor (#318).
4. Add selected-row CSV export (#315), document keyboard shortcuts (#316), and atomic tagging of selected documents (#317).
5. Make collection processing coverage and quality inspectable before relying on search or export results (#320).
6. Add consistent query snapshots (#321) and a workspace for browsing and acting on their exact membership (#322).
7. Add an exact-version document inspector (#323), verified text reader (#324), exact-version page images (#326), and page navigation (#327).
8. Write verified document bundles (#328), expose export preview and download (#329), and let verification finish while retiring stale page jobs (#331).
9. Decode email MIME into verified metadata (#309), publish exact attachment child documents (#330), and import MBOX or Gmail Takeout ZIP archives with resumable receipts (#332).
10. Add dedicated email review, verified email PDFs and batch exports with explicitly selected attachment sets. Build on the existing document and export services rather than introducing a separate bundle engine.
11. Add durable collection snapshots, resumable collection-package imports, supplied-text indexing and load-file parsing while preserving original identifiers and source relationships.
12. Add optional Bates stamp PDF exports and load-file exports, with verified page counts, stable allocations and preserved unstamped originals. Expose Bates stamp controls through the web, CLI and MCP.
[...]

PR #340, opened 2026-09-13
feat(timeline): index document dates from retained evidence

salmonumbrella:
## What changed

Docbank now derives document date claims from retained source metadata and provenance, keeping each claim tied to its exact content version. Dates retain their original precision and timezone information, and conflicting claims remain separate.

The daemon and embedded Go API expose rebuild progress and coverage. Backups preserve the source evidence and rebuild the timeline during restore; valid documents that exceed indexing bounds remain recoverable with explicit unavailable coverage.

## Why

Document dates come from different sources and can disagree. Keeping their evidence and uncertainty makes those dates usable without turning an inferred date into an asserted fact.

## Usage

Read current-file coverage with `GET /api/v1/timeline/coverage`. Start or replay a rebuild with `POST /api/v1/timeline/rebuilds` and a canonical UUIDv4 `operation_id`, then read its receipt at `GET /api/v1/timeline/rebuilds/{operation_id}`. These routes use the daemon’s existing authentication.

Embedded callers use `Vault.DocumentEventCoverage` and `Vault.RebuildDocumentEvents`.

Refs #333


wesm:
rebasing

PR #341, opened 2026-09-13
feat(loadfile): add package codecs and deterministic identity

salmonumbrella:
Add the internal parsing foundation for load-file packages: Concordance DAT, RFC 4180 CSV, two declared OPT column layouts, strict character decoding, validated column mappings, and deterministic package fingerprints. Imports need these codecs to preserve what the sender supplied before Docbank assigns document identities or stores content.

DAT and CSV preserve source column order, duplicate headers, row positions and raw values. LF, CRLF and CR record endings are accepted; blank rows follow the column-count diagnostic path. Malformed content has a separate error from an invalid profile. Declared headers and date formats are authoritative, date/time pairings must identify distinct mapped columns, and named timezones such as EST remain unresolved with a warning. All three formats stream records; their collecting helpers stop at 4,096 records or pages.

Manifest identity covers supplied source-file digests, records, page references, volumes and family metadata using canonical JSON. The model can carry metadata supplied by callers; these parsers do not assemble families, resolve sidecars, apply column mappings or combine paired time fields. Package preflight and import commands remain separate work.

Refs #333


wesm:
rebasing

wesm:
working on fixing the roborev unicode issue so we get reviews

wesm:
Rjected finding

PR #342, opened 2026-09-13
feat(transfer): add portable contract and local verification

salmonumbrella:
## What changed

Docbank can verify a Msgvault transfer directory, ZIP, or legacy JSONL export locally. Reports check package integrity and preserve the distinction between native transfer evidence and the older format’s missing authority.

The public Go contract provides canonical records, bounded validation, and lossless date evidence for producers and consumers. Verification does not open a vault or start a daemon, and an integrity pass does not authenticate the producer.

## Why

A portable export needs a shared contract and a way to catch malformed records, broken references, and mismatched bytes before import. Older Msgvault exports also need an explicit archive binding without inventing timestamps, sequence numbers, or evidence they never contained.

## Usage

```bash
docbank transfer verify ./msgvault-transfer.zip --json
docbank transfer verify ./messages.jsonl --archive-id mva_example_01 --json
```

This delivers roadmap slice DB-45a: the contract and local verifier. Vault import remains a separate slice.

Refs #333


wesm:
rebasing

PR #343, opened 2026-09-13
Show format support and provider qualification in CLI and APIs

salmonumbrella:
`docbank formats` and matching HTTP and embedded APIs show the running binary's capabilities for all 52 catalog formats. Operators can distinguish detection, retention, metadata, expansion, text, pages, and transcripts before processing; a recognized extension alone cannot establish those capabilities. Lookups also distinguish pending formats from unknown inputs.

All three surfaces use the snapshot captured after providers are configured. For example, PDF detection, retention, and metadata are qualified by local fixtures, while a configured provider's declared text support is unqualified. No shipped rendition provider is qualified for text, pages, or transcripts yet; the sole provider qualification exercises a synthetic PDF-to-Markdown test provider. Archive expansion remains unsupported.

Qualification requires executable fixtures tied to the active implementation identity or exact provider descriptor. Detection and retention identities are readable version labels, not source-code hashes. Inventory reads do not start providers or perform processing.

This adds inspection without changing stored vault formats or adding decoders. MCP binding and the shared operation registry remain separate integration work. Refs #333.


wesm:
rebasing

PR #345, opened 2026-09-13
Preserve timed transcript evidence across search results

salmonumbrella:
Transcript search hits now retain the matched recording interval across lexical, semantic, and hybrid search. A cue from zero to one second returns `time_span: {"start_ms":0,"end_ms":1000}` through the embedded API and daemon HTTP responses, including CLI JSON and the web client model.

Canonical timed transcript artifacts retain ordered text, optional speakers, and half-open millisecond intervals. Adjacent and overlapping cues stay distinct. Generated timed Markdown passes front-matter validation, while untimed transcripts and direct-file embeddings omit timing.

Search derives spans from the exact retained build and input generation, with current-version and scope checks. Each vault caches verified mappings for up to 100,000 inputs; cold lookups still read the generation and evidence artifacts. An unreadable or corrupt artifact drops the affected semantic hit and marks the report `truncated`. Search continues through later ranked candidates until it fills the usable result limit or exhausts the scoped candidates. No storage-format change is required.

The web change adds timing to the response model. This capture shows the current search interface, served by this branch's real daemon using a temporary synthetic vault.

Search results from a temporary synthetic vault

Provider execution, acquisition, timeline integration, and MCP propagation remain separate work. Refs #240.


wesm:
reviewing

PR #346, opened 2026-09-13
Retain supplied recordings and process transcripts through the daemon

salmonumbrella:
Retain supplied WAV and MP3 recordings with their original bytes, exact revisions, and independent caller occurrences. `docbank media` and the matching embedded and authenticated HTTP APIs support retention, transcript import, status, and explicit processing retries.

Retention starts no provider work. Processing uses existing consent and stays bound to the selected transcript, source revision, and original authorization through restart. Revoking an occurrence removes its access without changing another caller’s occurrence. Backup preserves the retained evidence and history; restored work requires fresh consent.

Synthetic WAV and MP3 consumers exercise transcript processing, exact rendition reads, and phrase search. Restart coverage also proves successful embedding completion and rejection of revoked authority.

Remote acquisition execution and the indexed media population integration in Task 8 / DB-14 remain separate follow-ups.


wesm:
rebasing

PR #350, opened 2026-09-14
feat(pdfstamp): stamp verified Bates PDFs

### LINKED ISSUE #349: Add a verified Bates PDF stamp engine
## The situation

Produced document packages need physical Bates labels that match their exact per-page assignment. The package work needs a small PDF transformation boundary it can call after allocation without letting the PDF engine choose, reorder, or replace labels.

## What works today, and why it is not enough

The repository already depends on pdfcpu, but calling its watermark API directly does not establish a stable recipe, prove label placement across cropped and rotated pages, reject an overflowing sequence, or keep partial output private when verification fails.

## Ask

Add a standalone `bates-stamp/v1` PDF engine that accepts an explicit label for every source page, stages and verifies the derivative, preserves the source bytes, and qualifies pdfcpu v0.15.0 with independently rendered synthetic PDFs. Keep allocation, export wiring, HTTP/CLI controls, and daemon worker supervision in their owning integration slices.

salmonumbrella:
## What changed

- Added a canonical `bates-stamp/v1` recipe pinned to pdfcpu v0.15.0 and an explicit label for every source page.
- Stamping now rejects incomplete or substituted labels, sequence overflow, clipping, malformed input, unsupported fonts, and existing watermarks unless restamping is requested. Output stays staged until page count and every declared label verify, with a 512 MiB publication ceiling.
- Qualified Letter, A4, cropped, rotated, mixed-size, hostile-input, and all nine placement cases with Poppler 26.09.0 renders in both SQLite modes. Normal tests remain portable when Poppler is unavailable.
- Kept the engine synchronous and isolated. The later daemon/export integration must run it in a bounded supervised worker before publishing package bytes.

## Why

The package exporter needs a deterministic PDF transformation it can place after Bates allocation. This gives it a verified derivative and exact page map while preserving received labels and source PDFs, without coupling the engine to unfinished allocation or export APIs.

## Usage

No usage change.

Closes #349


wesm:
Validated the reported failures and pushed fixes in `63971825`.

- Scanned PDFs and source forms containing the same text now stamp successfully. Verification checks the new label only.
- Source layers cannot hide the new label. Their existing visibility settings stay intact.
- Restamping replaces recognized old stamps, including when the label stays the same.
- Rotated and cropped pages keep their original content and displayed size. Tests compare source pixels before and after stamping.
- Stamping ignores host configuration. Recipes reject characters the engine rewrites, and a test checks the linked engine version.

Review also caught shared page resources, blank pages, and split content streams; those cases now have fixes and regression tests.

The complete Go suite passes in both SQLite modes, including independent Poppler 26.09.0 rendering checks. Lint and `prek run` pass.

Scope clarification: duplicate-stamp detection recognizes pdfcpu watermark artifacts. Arbitrary Bates text from Acrobat or other tools remains source content. No Acrobat-produced fixture was available for that specific claim.

wesm:
Confirmed both ways the label could be hidden, and pushed the fixes in `fbc24189`.

- Unbalanced drawing commands could let source clipping or transforms hide the stamp. Source drawing commands now run inside a separate PDF form so they cannot change how the label is drawn. Independent Poppler renders show the label in the expected position for each reproduced case. Ordinary balanced clipping and transforms already worked.
- An opaque annotation covered the label even though text extraction found it. **All annotated PDFs now require flattening before stamping, including restamping.** They fail before any output is written.

The new tests fail against the previous commit and pass with these fixes. Both full Go suites (CGO and pure Go), Poppler 26.09 rendering checks, lint, and pre-commit checks pass.

Rendering is a test qualification check, not a runtime step. At runtime, the engine checks the PDF structure; containing source drawing commands and rejecting annotations prevent the reproduced failures.

wesm:
Reviewed both findings and pushed `8685d4bd`.

**Tagged PDFs: confirmed and fixed by rejecting the input.** Stamping moved page content but left the accessibility references pointing to its old location. The engine now rejects tagged PDFs before writing output, including during restamping. The regression covers files with and without the optional flag that identifies them as tagged.

**Split strings: reproduced on a malformed PDF; stream joining stays unchanged.** A string split across two content streams does receive an extra newline. However, [the PDF specification, Table 30](https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/PDF32000_2008.pdf), requires stream boundaries to fall between complete tokens. A boundary inside a string violates that rule. Removing separators also merges otherwise separate drawing commands. The relaxed reader can accept the malformed example; this change does not add recovery for it.

The tagged-input regression fails on the previous commit and passes with the fix. Both full Go suites (CGO and pure Go), pinned Poppler 26.09 qualification, lint, and pre-commit checks pass.

PR #352, opened 2026-09-14
feat(loadfile): preflight received packages and read LFP page maps

salmonumbrella:
Inspect a received folder before importing it. `docbank package preflight` reads one DAT file and an optional OPT or LFP page map, checks document and page references, and returns counts, a content fingerprint, and diagnostics.

The authenticated daemon reads through a confined directory handle. It walks the folder once, bounds reads to inventoried file sizes, honors cancellation, and releases the handle afterward. Files that change size require another preflight after writing finishes. Local-path requests require loopback access and an explicit source path. Only saving the result holds the vault mutation lock.

Preflight saves a 24-hour receipt with paged diagnostics without creating documents. The package table has its own checked storage schema version. Explicit volume mappings must cover the discovered folders. Load files must sit inside a volume directory. LFP supports image and comment records.

```sh
docbank package preflight ./delivery \
  --profile dat-concordance-v1 \
  --encoding utf-8 \
  --json
```

For OPT files with page counts in column five, add `--page-map-profile opt-pagecount5-v1` (API: `page_map_profile`). Standard OPT remains the default; LFP defaults to `lfp-ipro-v1`. A profile that does not match the page-map format is rejected.

Use `--map mapping.json` for an explicit `loadfile-mapping/v1` mapping. Package import remains separate work.

Refs #333


wesm:
reviewing

wesm:
Addressed the review findings in `895e1ae7` and `573045d7`.

- Gave the package table a new storage schema version and added checks that reject a missing or malformed table. Existing released-vault upgrade tests cover the new layout.
- Removed the unimplemented container endpoint and CLI fallback, the unused directory-handle registry, and unused helper APIs. A mistyped folder path now gives a directory error.
- Removed the 60-second preflight timeout and applied the existing loopback restriction. Inventory, LFP scanning, and hashing check cancellation.
- Reused one folder inventory, replaced repeated page-count scans with a lookup, and limited the vault mutation lock to saving the result.
- Reject incomplete volume mappings. Malformed data, unsupported LFP commands, missing roots, and size-limit failures now return correctable 4xx errors.
- Connected LFP page maps to preflight, removed the unused writer, and shared record hydration with field mapping.
- Added checks for cancellation, oversized inventories, hourly receipt expiry, invalid diagnostic cursors, and schema rejection. Root-level load files now clearly explain the required volume-directory layout.

Validation: full Go suite with CGO and pure-Go SQLite; focused race checks; lint (0 issues); strict docs build; Windows loadfile test cross-compilation; and `prek run`.

wesm:
The expiry finding was valid and is fixed in `8deface7`. Saved preflight results and their diagnostics now stop being readable at the expiry time, even if background cleanup has not run. A test checks both sides of the exact 24-hour boundary while the expired row still exists. The hourly cleanup test also proves the row is deleted.

Squashed the branch to one commit and rebased it onto current `origin/main`, preserving salmonumbrella as author and crediting Wes as coauthor. The earlier review fixes are included.

Checks passed after rebasing: the full Go suite with both CGO and pure-Go SQLite, build, lint, strict docs build, and `prek run`.

wesm:
Fixed both findings in this update:

- The API description now includes the package request and response fields, UUID path values, diagnostic pagination, and error responses. A contract test checks those declarations.
- Canceling a preflight now stops OPT scanning even when the remaining rows are malformed. A regression test covers that case.

Squashed and rebased onto main, preserving salmonumbrella as author. The package table now uses schema version 16 to follow the person-authority schema added on main.

Both full Go test suites (CGO and pure-Go SQLite), the Go build, release-script tests, lint/prek, and strict docs build pass. Force-pushed as `f7758b96`.

wesm:
Confirmed and fixed both findings in `c1108305`.

- Preflight now checks the remaining shared memory budget before mapping allocates lists, family links, or extra fields. This stops large semicolon-separated fields from expanding past the limit before rejection.
- Every record with an empty or unmapped document ID now produces a blocking diagnostic. Each diagnostic identifies the source file and row so the input can be corrected.

Regression tests cover allocation refusal, budget use across records, and empty or unmapped IDs through the HTTP API. Both full Go test suites (CGO and pure-Go SQLite), build, lint/prek, and strict docs build pass.

wesm:
Confirmed and fixed in `5bc0da4a`. Preflight now uses the existing CSV parser when the caller selects `csv-rfc4180-v1`, so quoted multiline fields are normalized correctly in the saved manifest.

The HTTP regression test checks embedded CRLF, commas, escaped quotes, and the following record in the stored manifest. It failed before the fix and passes now. Both full Go test suites (CGO and pure-Go SQLite), build, lint/prek, and strict docs build pass.

PR #355, opened 2026-09-15
Add person and custodian authority core

salmonumbrella:
Docbank's storage core now retains a person directory, source identities, and custodian claims for collections and exact document versions. Matching names remain separate until explicitly merged, and splits move only selected identities and assignments.

External identities stay scoped to their Msgvault archive, and alias chains can follow later merges and splits. Existing event actor keys remain readable alongside scoped handles and external UID keys. Revision checks reject stale changes. Retiring a person keeps their custodian labels visible as unresolved claims.

Media retries preserve admission order when clock readings tie, including through backup and restore.

This PR provides the storage foundation. Person browse routes, resolver publication, package scope, and transfer consumption remain separate work. **Metadata export and backup refuse vaults containing person authority** until the JSONL format can preserve it; they cannot silently produce an incomplete backup.


wesm:
reviewing

PR #356, opened 2026-09-15
Add reviewable person match candidates

salmonumbrella:
Docbank can retain ambiguous person matches for review. Operators can link a match, reject it, or create a person. Reprocessing the same evidence preserves its decision; new evidence creates a separate item. The queue reports when its 10,000 open-item limit is reached.

Links apply only to captured document versions that still exist. Pruned evidence stays in the decision history. Merging duplicate people combines matching assertions, preserves their history, and refuses opposing assert/suppress decisions. Linked decisions follow the surviving person. Merges and retirement supersede affected open suggestions; retirement keeps earlier assertions.

This adds the storage foundation, with no new review interface. Metadata export and backup refuse vaults containing candidate history until the export format can preserve it.

Cap deployment probes also report timeouts as provider unavailability when a response ends without a read error.

Builds on #355 (merged).


wesm:
rebasing

PR #357, opened 2026-09-15
Retain timeline actor evidence

salmonumbrella:
Docbank keeps timeline actors tied to their source evidence, including calendar organizers and creators without dates. Undated claims remain attributable without inventing timestamps.

Events, actors, and primary dates are stored together. Older event records remain readable, and reopening a vault checks the complete event schema. This adds storage support, with no new person matching behavior or interface.

Builds on #356 (merged).


wesm:
rebasing

PR #358, opened 2026-09-15
Publish deterministic document-person attribution

salmonumbrella:
Docbank can publish document–person links from retained actors, custodian assignments, and operator decisions. Exact identifiers reuse or create people; name-only matches stay in review, and display-only actors remain unresolved. Reused identifiers remain unresolved, with active people suggested for review. Long labels are shortened without changing the source evidence or identity keys.

Publication rejects stale inputs and reuses immutable results on retry. Reads hide outdated links until they are rebuilt. Changes to collection membership or current content select the affected document for rebuilding. Oversized inputs record unavailable coverage without keeping stale links.

Person resolution creates records that backup and metadata export cannot yet preserve. Those operations refuse vaults containing person records, so automatic scheduling and public rebuilds must wait for person JSONL support.

Builds on merged #357. This adds storage and resolver support, with no new interface.


wesm:
rebasing

wesm:
I reproduced the five functional findings and fixed them:

- New senders no longer invalidate every previously resolved document. Operator edits still invalidate earlier resolutions.
- Removed the unused cached person counts. They became stale after replacement or deletion, and no product code reads them. Counts can be computed from current documents when a reader is added.
- A matching name remains a review suggestion, even when only one person matches. Actors without an identity key remain unresolved instead of failing the document.
- Failed input preparation no longer tries to write status using incomplete inputs. Size-limit failures retain the version and authority checks needed to record unavailable coverage.

I also removed unused fields and the inactive dirty queue, made the preparation method's name reflect its writes, simplified backfill error handling, and changed the external-identity test to publish a valid event instead of disabling an immutability trigger. The external-identity scan stays as-is until there is a measured reason to index it.

The PR description now states the backup constraint: person resolution can create records that backup and metadata export cannot yet preserve. Automatic scheduling and public rebuilds must wait for person JSONL support; the existing refusal remains in place.

Updated in [cae53033](https://github.com/kenn-io/docbank/pull/358/commits/cae53033ced6305a3d4e9ba277a5c95cfec425b3). Focused regressions, the full test suite in both SQLite modes, lint (0 issues), and pre-commit checks all passed. The PR remains one commit on the latest main, with the original author and message preserved.

PR #359, opened 2026-09-15
Run and report document-person rebuilds

salmonumbrella:
Docbank can now rebuild document-person attribution through the daemon or embedded API and report whether current files are published, pending, failed, or unavailable. Repeating an operation ID returns the original receipt, while the worker resumes running builds and waits through late dirty versions and retryable failures.

The authenticated HTTP surface exposes rebuild submission, status, and coverage. Coverage also reports unresolved actors and custodians, plus candidate-queue pressure, without exposing person evidence.

This is Task 11 of DB-55 and builds on #358. Task 12 will add deterministic authority backup and restore.


wesm:
rebasing

PR #361, opened 2026-09-15
Keep person coverage truthful through lifecycle changes

salmonumbrella:
People coverage now stays accurate when documents are trashed or restored, current versions change, old versions are pruned, source evidence is withdrawn, or email-derived evidence is purged. For example, trashing a document removes it from active person rollups while retained-item coverage still counts it; restoring it brings the rollup back. Purging an email-derived source removes inferred relationships while preserving manual person assignments and the durable suppression that prevents the same evidence from being reimported.

This stacks on #360. It also bounds derived actor labels, avoids republishing the whole corpus when a person is created automatically, keeps metadata export valid after referenced versions are pruned, and prevents custodian-shaped fields from entering retained source metadata.

The people directory and CLI usage page stays with the later surface PR because those commands are not present in this branch.


wesm:
rebasing

PR #362, opened 2026-09-15
feat(docling): map qualified ASR track evidence

salmonumbrella:
## What changed

- Add a qualified Docling Serve path for PCM WAV and MP3 recordings, pinned to the service schema and versions exercised by synthetic spoken fixtures.
- Convert Docling's track-source timing into canonical transcript segments while preserving overlaps and optional speaker labels. Missing, ambiguous, regressing, or out-of-range timing fails explicitly.
- Inspect the sealed recording locally before any provider request, enforcing the qualified container, codec, size, and duration bounds. Existing document conversion behavior stays unchanged.

## Why

Docling returns speech timing inside each text item's track source, while Docbank previously handled only ordinary document output. Mapping any other shape would risk publishing invented or unbounded transcript evidence, and sending an unqualified recording would disclose bytes before Docbank knew it met the supported contract.

## Usage

No usage change.

Refs #333. Daemon configuration and supplied-input selection remain separate follow-ups after #346.


wesm:
looking

wesm:
Pushed fixes for the review in `20999635`:

- Requests that cannot retain a transcript now stop before reading or uploading the audio.
- Approval now covers the qualified service/schema versions and transcript limits. An earlier approval cannot be reused after those settings change.
- Final timestamps may extend up to 20 ms past the measured duration. The retained timestamp is clamped to the recording's end, including the floating-point rounding case.
- Blank cues no longer discard useful speech. Verbose provider JSON uses the response limit; the saved transcript keeps its separate 16 MiB limit.
- The client reuses the upload's validated audio inspection. Withheld filenames consistently become `document.wav` or `document.mp3`.
- Speech transcription now has its own client and shares the HTTP task flow with document conversion. Duplicate checks and settings that only accepted one value were removed.

Passed the full Go suite in both SQLite modes, Docling race tests, lint, `prek run`, and the strict docs build. These were local checks; I did not rerun live provider qualification.

Overlong speaker labels still reject the result; silently skipping those cues would lose speech. The separate daemon integration still needs to match the worker's text limit to the ASR profile.

wesm:
Windows follow-up pushed in `5697e120`.

The failed test gave vault shutdown five seconds after releasing a content reader. Cleanup subsequently completed; the Docling tests passed. Closing the vault also waits for workers and storage, so that deadline tested shutdown speed as well as reader ownership.

I removed the five-second deadline from the range-reader and email-reader tests. They still use real vaults, check that an open reader prevents closing, and require successful shutdown after releasing the reader. The suite timeout still catches permanent hangs.

Local checks passed against the exact merged source that CI tested: 30 repetitions of both tests under the race detector, 30 with pure-Go SQLite, and the full root-package suite in both SQLite modes. The branch's root-package suite and `prek run` also passed. Windows execution of this patch remains unverified until the new CI run completes.

PR #364, opened 2026-09-15
feat(loadfile): serialize DAT, CSV, and OPT outputs

salmonumbrella:
Docbank can now write deterministic Concordance DAT, CSV, and Opticon OPT files. Exports preserve declared column and page order and can exceed 4,096 rows without repeating headers or encoding markers.

Writers check all rows and encodings before emitting output. They format output twice to keep memory bounded. Disk errors can still leave partial files; callers must write to a temporary file and rename it after writing and closing successfully.

CSV preserves raw text, including formula-like values. The web spreadsheet download has its own escaping policy. OPT omits source-page, boundary, and rotation metadata because the format cannot store it; callers retain those separately.

This is the internal serialization layer. Package assembly and export profiles remain in #328; LFP export is separate in #352. There are no new CLI commands.

Refs #333


wesm:
looking

wesm:
Fixed the serializer findings in c9ea0c55:

- DAT, CSV, and OPT exports can exceed 4,096 rows without repeating the header or encoding marker.
- DAT preserves carriage returns and rejects column names containing newlines or its newline marker. Profiles with no newline marker or a carriage-return marker work in both directions.
- CSV rejects embedded CRLF values because its reader would silently change them to LF. Lone carriage returns remain supported.
- OPT rejects literal backslashes in paths, which its reader would otherwise change into slashes.
- Writers reject leading content that would be mistaken for an encoding marker, including the UTF-16LE/NUL case.
- All rows and their encoding are checked before writing any output. Invalid data leaves the destination untouched. File callers must still discard partial output after an I/O failure.

Also removed the unused export registry, shared profile checks and encoding code, simplified OPT column handling, and removed checks for fields OPT does not use. Standard DAT and OPT rows now use CRLF endings.

Regression tests, the full Go suite in both SQLite modes, vet, lint, and `prek run` pass.

wesm:
**Headerless DAT: fixed in 09dc6d52.** Column names are now checked for newline characters only when the file includes a header. A regression test reads, writes, and reads headerless DAT with CR, LF, and the newline marker in its column names, preserving both the file contents and column metadata.

**CSV formulas: no change to this raw serializer.** This internal load-file writer has no production callers yet. Its job is to preserve supplied field text and column names. Rejecting formula prefixes would also reject ordinary values such as negative numbers; adding an escape prefix would change the exported data.

The web app's **Export page CSV** feature uses a separate spreadsheet exporter. I exercised that helper and confirmed it prefixes formula-leading text, including text preceded by whitespace. I also verified that the raw writer preserves formula-like headers and values on read-back.

Opening raw CSV in a spreadsheet can still evaluate its contents, as [OWASP describes](https://owasp.org/www-community/attacks/CSV_Injection). I left spreadsheet-specific escaping at the spreadsheet export boundary so load-file values remain unchanged.

The full Go suite passes in both SQLite modes, and `prek run` passes, including lint.

wesm:
Pushed `0a1f7282`.

The linked Windows ARM64 job failed in `TestMediaStatusSeparatesLatestAttemptFromSuccessfulCoverage`. Its load-file tests passed.

I reproduced the wrong-result pattern with a frozen clock: two attempts received the same timestamp, so sorting by random operation ID selected the older success. New media operations now receive strictly increasing admission timestamps for each caller and source, within the existing database transaction. This preserves retry order through metadata backup and restore.

The new regression test checks the newest attempt and the previous successful coverage before and after restore. It passes in both SQLite modes. The original failing test also passed 20 local runs in each mode. Full Go suites, lint, and commit checks pass locally; the updated Windows job has not been verified.

The repeated CSV finding is still not applied to the raw load-file writer. Changing or rejecting formula-leading text would alter or reject valid source values. The web app's spreadsheet download uses its separate escaping policy. I added this distinction to the `WriteCSV` documentation; this does not claim that opening raw CSV in a spreadsheet is safe.

wesm:
rebasing

Issue #351, opened 2026-09-14
Roadmap: add local unit counters for non-PDF Mistral formats


rodboev:
Docbank's `document/mistral` package processes only PDF and PPTX because no other format has a locally enforced unit bound. Mistral accepts 18 format families, but the upload path rejects anything it can't measure before sending. That blocks kenn-io/msgvault#608.

The work lands as two lanes plus one library-backed counter. Renderable formats convert to PDF locally through a sandboxed LibreOffice runtime, get counted by the existing PDF page counter, and upload that exact PDF, so the counted bytes and the uploaded bytes are the same object. Text formats stream through stdlib record and line counters. XLSX keeps a direct-upload sheet counter built on excelize. #175 tracks the broader ask; this issue tracks the concrete delivery.

## Outcomes

- XLSX, DOCX, and the text families pass through the Mistral upload path with locally enforced unit bounds.
- Renderable formats upload the locally counted PDF, so the pre-upload bound is exact instead of a guess about provider pagination.
- Formats without a counter or conversion route stay blocked. The guard test catches regressions.

## Delivery sequence

1. PPTX slide counter.
   PR: #353

2. XLSX sheet counter built on excelize.
   PR: #378

3. Format-neutral render-to-PDF pipeline with a pinned, sandboxed LibreOffice runtime.
   PR: #382

4. DOCX through the render pipeline, counting and uploading the generated PDF. Closes #363.
   PR: https://github.com/kenn-io/docbank/pull/385

5. Text-family streaming counters for TXT, Markdown, CSV, JSON/JSONL, YAML, source code, and EML/MSG.
   PR: #379

6. Render lane for DOC, ODT, RTF, PPT, XLS, ODS, and XLSX.
   PR: https://github.com/kenn-io/docbank/pull/391

Steps 2 and 5 are independent. Step 3 gates step 4, which gates step 6's render entries.

## Design constraints

- Counters use established libraries or stdlib. excelize replaces bespoke OOXML parsing for XLSX; pure-Go DOCX parsing lost because pages are a rendering concept and unioffice's license excludes it.
- The renderer is an optional trusted runtime. The pipeline launches an immutable copy of the SHA-256-verified executable, limits filesystem access to the per-call work directory, disables outbound network, and strips or rejects external relationships before conversion.
- The authorization binds to exact bytes via SHA-256. Conversion receipts record both the original and generated hashes.
- A file that can't be counted or converted is rejected, never sent.
[...]

PR #365, opened 2026-09-15
feat(mistral): authorize xlsx with local sheet bounds

rodboev:
XLSX workbooks can use Mistral OCR with locally enforced sheet limits and a fresh capability manifest, so callers can submit the original workbook instead of exporting it to PDF.

The counter follows workbook relationships, counts hidden and empty worksheets, checks the limit before upload, and rejects chart-only or Strict workbooks it cannot count. Provider counts must match. It extends the [PPTX approach](https://github.com/kenn-io/docbank/pull/353).

Refs #351


mariusvniekerk:
This feels like we should probably have a library for this. I don't think we should build our own XLSX parser from scratch. This seems kind of excessive.

rodboev:
Yes good point, reworking it on top of excelize.

rodboev:
Reworked this on excelize in https://github.com/kenn-io/docbank/pull/378.

PR #367, opened 2026-09-15
feat(docxpdf): convert DOCX locally to PDF for bounded OCR

### LINKED ISSUE #363: DOCX page counter needs local conversion to measure pages before upload
The Mistral upload path cannot enforce a pre-upload page limit for DOCX files. PPTX and XLSX have structurally countable units (one ZIP entry per slide or sheet), but DOCX pages are a rendering concept with no structural equivalent. The `docProps/app.xml` metadata can declare a page count, but nothing prevents it from being wrong. A document declaring one page could contain ten page breaks and render to eleven pages. The system would send all eleven pages to Mistral before discovering the mismatch, wasting provider cost on a file it rejects after the fact.

Convert the DOCX to PDF locally before counting, following the precedent in #293 (CSV to PDF conversion). The local converter produces exact page counts from rendered output. Register the DOCX counter with the same local-exact bound method PPTX and XLSX use. This unblocks #351 slice 3 and downstream kenn-io/msgvault#608.

## Outcomes

- A DOCX file exceeding `MaxUnits` is rejected before any bytes reach Mistral.
- The local page count matches the provider's reported `pages_processed` for well-formed documents.
- Files that fail conversion are rejected rather than sent unbounded.

## Design constraints

- The converter (Docling) is already in the dependency tree. #293 proves the conversion path works for CSV.
- Conversion fidelity matters: if the local converter paginates differently than Mistral, the post-upload exact-match check rejects valid files. Test against real DOCX samples from Word and LibreOffice.
- The conversion runs once per upload, purely for counting. Mistral still receives the original DOCX for its own processing.

## Adjacent

- #351 tracks the three OOXML counters. Slice 1 (PPTX) shipped in #353. Slice 2 (XLSX) follows the same structural-count pattern. Slice 3 (DOCX) is blocked on this issue.
- #175 is the parent ask for bounded non-PDF authorization.
- kenn-io/msgvault#608 is the downstream consumer, blocked on #351.
- #293 shipped bounded CSV to PDF conversion and is the implementation precedent.

rodboev:
Go applications can convert DOCX to PDF with local LibreOffice, then count and send that exact PDF through existing Mistral bounds. A receipt links the original document to the generated PDF.

Manual PDF export remains an alternative. A private conversion profile and finite limits keep failed or excessive conversions from yielding uploadable output. LibreOffice is an optional trusted runtime. This adds a reusable library, following [CSV conversion](https://github.com/kenn-io/docbank/pull/293); application integration remains separate.

Closes #363


mariusvniekerk:
Shouldn't we use a library for some docx parsing? This feels like something we can import for a lot of functionality.

rodboev:
Yeah you're right, and the renderer integration probably needs rethinking too. Excel is more straightforward since there's excelize, but I'm not sure how good our options are on the docx/pptx side. I think we might need to use headless LibreOffice there.

Closing this for a rework.

rodboev:
Reworked this as a format-neutral renderer library in #382. LibreOffice now owns DOCX import and PDF export inside a private-root sandbox, with flat-ODF admission between the two stages. The DOCX routing work stays in #363.

PR #368, opened 2026-09-15
fix: match collections by ID instead of assuming positional order

rodboev:
Windows arm64 can assign two ingests identical timestamps. The UUID tie-breaker then changes their sort order, but the test assumed positional stability. This looks up collections by ID so assertions hold regardless of ordering.


mariusvniekerk:
This feels like it's solving the problem at the wrong level. If we rely on ordering, we should do the ordering outside the test.

rodboev:
Thanks, good catch. Closing and redoing properly. This was my fault, vet refused the fix.

PR #378, opened 2026-09-16
test(mistral): keep XLSX probes from granting upload authority

rodboev:
Keep direct XLSX production uploads blocked until provider billing can be bounded before upload. The existing one-cell probe cannot distinguish one unit per worksheet from one unit per rendered page; a mismatch detected after upload cannot prevent charges.

Regression tests verify that a successful XLSX extraction probe grants no upload authority and that manifests claiming an exact worksheet-based bound are rejected. Documentation records the existing worksheet and chartsheet observations and their limits. Direct XLSX support still needs representative provider evidence and enforceable counting rules.

Refs #351


wesm:
Changed the sheet counter so it no longer decodes and repairs every cell before upload. Two small regression workbooks previously allocated about 111 MB and 40 MB; both now allocate less than 1 MB. The tests cover both reported input shapes without recreating the multi-gigabyte load.

I did not use the proposed comparison of sheet totals alone. It accepted a missing worksheet reference and a chartsheet masked by an unused worksheet, while rejecting a valid worksheet with a different ZIP filename. The counter now uses the workbook's sheet list, checks references without reading rows, and rejects Excelize's recognized non-worksheet parts, including unused ones. I also added explicit error expectations to the test table and documented the memory cost of large workbooks.

The billing question remains unverified. A one-cell probe does not establish whether a long worksheet is charged as one unit or several rendered pages. The docs now state that limitation and the two-sheet control needed to check it; a mismatch after upload cannot prevent charges. I did not run a paid Mistral probe.

The full Go suite passed with both SQLite implementations. Lint, `prek run`, and the strict docs build also passed.

wesm:
Changed the PR to keep XLSX production uploads blocked. The documentation caveat I added earlier did not enforce a spending limit before upload.

- **Billing evidence:** confirmed that the one-cell probe granted upload authority without distinguishing worksheets from rendered pages. Successful XLSX extraction now remains unbound, and manifests claiming an exact worksheet-based bound are rejected. No new live-provider billing evidence is claimed.
- **Renamed chartsheets:** reproduced the bypass. Moving a chartsheet under a worksheet path, while retaining its chartsheet metadata, passed preparation and counted as a worksheet. The sheet counter and Excelize dependency are now removed because they cannot authorize production uploads without the missing billing evidence.
- **Duplicate sheet IDs:** the reported undercount is incorrect for Excelize v2.11.0. `GetSheetList` keeps both workbook entries; the probe returned two sheets and two local units even with duplicate IDs. `GetSheetMap` collapses them, but the counter did not use that API. No speculative duplicate-ID parser was added.

The PR now adds authorization regression coverage and documents why direct XLSX processing remains deferred. Its title and description have been updated to match.

The full Go suite passed with both SQLite implementations. Lint, `prek run`, and the strict docs build passed.

PR #379, opened 2026-09-16
feat(mistral): authorize fourteen text formats

rodboev:
Docbank can authorize fourteen text formats for Mistral after a successful authenticated capability probe. Text uploads are limited by bytes; lines and records are no longer treated as pages. A CSV with more than 5,000 rows is therefore not refused solely because of its row count.

Docbank checks the response: returned pages must be positive, match `pages_processed`, and stay within `MaxUnits`. These checks cannot prevent processing or charges already incurred. Applications that require a pre-upload page or spending bound must not use this text route. This replaces the text-counter approach described in #351; PDF and PPTX keep their existing checks.

Capability manifests use schema v4. Operators must rerun the authenticated probe and replace v3 manifests before processing any format, including PDF and PPTX.

Refs #351


wesm:
Pushed 2f6569f1.

Changed the text formats to use a page limit checked after Mistral responds. Lines and records do not tell us how many pages Mistral will process, so the local counters rejected useful files without establishing a provider page bound. Files with more than 5,000 lines or records can now proceed when they fit the byte limits.

Removed the text counters and their duplicate parsing. Preparing a 1.5 MB synthetic YAML file now allocated about 3.1 MB in the local check, down from about 111 MB. Text detection still has its existing 50 MiB ceiling.

The response must contain positive page counts, agree with `pages_processed`, and fit `MaxUnits`. This is explicitly a response check: a rejected response may still incur provider charges. PDF and PPTX keep their existing checks.

Updated the tests and documentation, including the upgrade requirement: all schema-v3 manifests must be replaced by rerunning the authenticated probe, even for PDF/PPTX. This intentionally replaces the text-counter approach in #351.

The full Go suite passes with both SQLite drivers. Lint, vet, the strict docs build, and pre-commit checks pass. No live provider probe was run.

PR #381, opened 2026-09-16
feat(config): enable Docling ASR for supplied audio

rodboev:
Docbank can transcribe supplied WAV and MP3 recordings without provider transcripts. Operators select a configured Docling rendition in the existing media flow. Generated evidence stays marked as generated.

The media lane and qualified ASR adapter already retain and process supplied media. Runtime configuration adds the endpoint, transport, egress policy, and credential reference. The disclosure fingerprint binds the endpoint and deployment to the portable profile. The daemon verifies the descriptor and discloses the destination before consent.

Operators can process original audio through the daemon instead of supplying a transcript or using an embedded caller. Runtime configuration provides that path. Remote recording acquisition and message-provider parsing remain separate.

Refs #240


wesm:
false positive

PR #383, opened 2026-09-17
feat(processing): bind manual recordings to remote references

rodboev:
Docbank can bind a manually imported recording to an existing remote reference. The source identity stays stable, each caller keeps its own occurrence, and retained bytes stay separate from transcript coverage.

The current workaround is to import the audio as a supplied source and track the link separately. This adds the association to media storage and the embedded and authenticated HTTP APIs, using the existing import and processing paths. It needs no new configuration. The client protocol advances from 65 to 66, so upgraded clients and daemons must use matching builds.

The fallback accepts WAV/MP3 originals, then captions or transcript inputs. Network acquisition and Loom/Cap adapters follow separately. Plain-text transcripts use existing explicit processing; caption retention alone does not claim transcription.

Refs #240


wesm:
rebasing

wesm:
Pushed the review fixes:

- Source matching is now separate from the processing plan, so selecting a recording does not change the plan fingerprint.
- Processing and reading a transcript require the recording's own source binding. The old hash-only fallback is gone.
- Canonical URLs treat an empty path as `/` and convert Unicode domains to punycode. Trailing-dot hosts remain distinct.
- Shared the duplicate storage code, simplified transcript lookup, made retry receipts use an exact source version, and removed all 19 marker logs.

The docs now explain that media retry selects the newest visible occurrence. Older recordings can still use ordinary document processing. They also state the CLI's canonical URL limitation and that acquisition planning uses configured origins.

Rebased onto main, which already includes the CLI test repair from #352. That unrelated repair is no longer part of this PR.

The full Go suites passed in both SQLite modes, along with lint, API contract checks, and the strict documentation build.

Issue #374, opened 2026-09-16
storage:pack and extract:plain-text stay 'running' from daemon start and block every mutation


Joi:
## What happens

Both background jobs go to `running` within a tenth of a second of daemon start and stay there. While they do, every mutation is refused with:

```
error: vault maintenance is running or queued; retry this mutation after it finishes: vault maintenance is busy
```

The vault has been read-only for 14.5 hours as a result. Reads through `ls` still work; `info` and `storage status` do not.

## Evidence

```
$ docbank daemon status
running: pid 840 at 127.0.0.1:7486 (0.14.0), up 14h28m17s

$ docbank jobs --json
{"items":[
  {"name":"extract:plain-text","status":"running","started_at":"2026-09-15T21:05:44.079849Z"},
  {"name":"storage:pack","status":"running","started_at":"2026-09-15T21:05:44.079975Z"}]}

$ docbank info
error: daemon error (500 internal): reading blob-store inventory: context deadline exceeded

$ docbank storage status
error: daemon error (500 internal): reading blob-store inventory: context deadline exceeded

$ docbank ls /zotero | head -2
SELECTOR  KIND  SIZE  MODIFIED              NAME
id:17162  dir   0     2026-09-13T21:40:52Z  245LAL23
```

Both jobs start at the daemon's own start time, so nothing requested them.

Neither job is doing any work:

- 8m47s of CPU consumed across 14h33m of wall clock, and 0.32s across a 20-second sample.
- Zero files anywhere under the vault directory modified in the last two hours (`find ~/.docbank -newermt '-2 hours' -type f` returns nothing). The last write was the nightly ingest at 03:40 local.

So this is not a long pack on a large vault. The jobs are wedged, and the maintenance gate they hold keeps the vault read-only indefinitely.

## Expected

A wedged maintenance job should not block writes forever. Any of these would fix the immediate problem:

1. The maintenance gate times out, or is released when a job makes no progress for some interval.
2. `docbank jobs` reports the operation id for these jobs, so `docbank jobs cancel <operation-id>` is usable. Right now `jobs` prints only name, status and start time, and `cancel` requires an id that nothing gives you — there is no documented way out except restarting the daemon.
3. The jobs do not register as `running` before they are actually running.

The timing-out `reading blob-store inventory` on `info` and `storage status` looks like the same underlying stall, and may be the more useful thread to pull.

## Environment

[...]

PR #384, opened 2026-09-17
fix(maintenance): bound the scheduled pack maintenance

rodboev:
Scheduled packing now requests cancellation when a pass reaches the configured interval. When the pass honors cancellation, the daemon releases the maintenance gate after the operation returns and retries remaining loose content on the next pass. A filesystem call that ignores context can still delay that return.

The startup text-extraction pass now enters the same mutation gate as its per-document writes and honors the daemon context while waiting. The long-lived pack and extraction jobs still report as running, and an explicit operator-requested pack keeps its existing operation lifetime.

This covers the scheduled maintenance and startup admission paths from the report. The change doesn't establish the cause or resolution of the separate inventory timeout and intermittent daemon-status delay. The report's observed commands and process samples supplied the scope and impact.

Refs #374


wesm:
reviewing

PR #385, opened 2026-09-18
feat(mistral): authorize docx through rendered pdf

### LINKED ISSUE #363: DOCX page counter needs local conversion to measure pages before upload

rodboev:
Go applications can authorize DOCX for Mistral when they configure renderpdf and provide PDF capability evidence. The client converts once, counts the generated PDF, enforces page and byte limits before upload, and retries with the same PDF bytes. Native DOCX remains unauthorized without this conversion policy.

Processor and rendition consumers use the same route. Conversion failures distinguish rejected source documents from renderer faults and policy limits. Processor results include the original source hash for every format and the uploaded PDF hash when conversion occurs.

DOCX results retain the original Word family. Rendition receipts identify both the original DOCX and uploaded PDF; `Usage.InputBytes` reports the uploaded PDF size rather than the DOCX size, matching provider accounting. This adds no daemon configuration or CLI operation.

Closes #363; Refs #351


wesm:
reviewing

PR #388, opened 2026-09-18
feat(processing): register self-hosted Cap origins

rodboev:
Operators can now register a self-hosted Cap deployment by its exact origin. Docbank recognizes share and embed links on that origin, keys each recording by its Cap video ID, and lists whether the deployment answered a probe under its configured network and TLS policy. Before this, such a link was a generic URL reference keyed by its full text.

Each `[media_origins.<name>]` entry names the origin, allowed address ranges, certificate pins, one credential binding, and the deployed Cap revision. Matching compares normalized scheme, host, and port exactly, so a lookalike host, another port, or an unregistered site with `/s/<id>` stays a generic reference. A request naming another origin's credential fails before anything is stored.

The probe calls Cap's documented developer API through the pinned transport, which refuses redirects. Cap publishes no version endpoint, so the origin listing reports the revision the operator declared next to the probe result. Recordings still report `access_required`, because Cap documents no export for received share links. The manual import from #383 remains the fallback.

This adds configuration and origin listing fields without a storage change. Embedded registration, byte download, Cap Cloud, and Loom remain separate work. The exact-origin rule follows the self-hosted Cap item in #240.

Refs #240


wesm:
rebasing

PR #390, opened 2026-09-18
feat(processing): acquire supported Loom recordings

rodboev:
Docbank can process an authorized Loom recording from exported MP4 and SRT files. It retains the video, preserves each caption's timing and provider label, and returns timed intervals from fenced lexical and auto search.

Recognition stays local. Docbank records a Loom reference as unsupported for automatic acquisition, then accepts the exact exported files. Imported captions use the local supplied-captions profile regardless of their retained origin label, matching the transcript path. The retained label stays unchanged; derived evidence records that the caption file was supplied by the caller.

Remote MP4 imports enforce byte, duration, pixel, frame, and format limits before publication, with the byte limit applied before staging. Caption processing rejects malformed SubRip or any cue ending beyond the measured recording duration, even by one millisecond; it preserves cue times without clipping or tolerance. Semantic and hybrid retrieval and automatic network acquisition remain unavailable for this path.

Loom references previously retained under the generic URL identity still replay to their original source, but a new occurrence for the same recording creates a separate Loom source.

Refs #240


wesm:
rebasing and reviewing

Issue #392, opened 2026-09-19
Evaluate Jev reranking for natural-language file search in the web UI and TUI


salmonumbrella:
@rodboev we should also add this to Docbank: https://github.com/kenn-io/msgvault/issues/877

Imagine typing something like “eating chocolate pie” into the web UI or TUI and quickly pulling up the relevant files, even when you don't remember the filename or exact wording. Being able to pick a result and find similar files would also be useful.

Docbank already has optional reranking providers tracked in #221. Can we evaluate Jev through that path and expose this flow in both search frontends? The goal is better relevance while keeping interactive search fast. Jev would rerank the retrieved candidates; natural-language retrieval and finding similar files still need to work end to end.

Related: #176, kenn-io/msgvault#877.

rodboev:
It looks like most of the reranking is already there from #221, it just isn't hooked up to search yet. I can take it in slices:

1. Hook the existing reranker into search (config, consent, API, CLI)
   PR: https://github.com/kenn-io/docbank/pull/395
2. Add Jev as a provider
   PR: https://github.com/kenn-io/docbank/pull/396
3. Measure Jev against ZeroEntropy and Cohere: quality, latency, cost
   PR: https://github.com/kenn-io/docbank/pull/405
4. Natural-language search and a rerank toggle in the web UI and TUI
   PR: https://github.com/kenn-io/docbank/pull/404
5. Find similar files
   PR: https://github.com/kenn-io/docbank/pull/394

The numbers from 3 should help with kenn-io/msgvault#877 too.

rodboev:
The five slices described here are now on `main`: [PR #394](https://github.com/kenn-io/docbank/pull/394), [PR #395](https://github.com/kenn-io/docbank/pull/395), [PR #396](https://github.com/kenn-io/docbank/pull/396), [PR #404](https://github.com/kenn-io/docbank/pull/404), and [PR #405](https://github.com/kenn-io/docbank/pull/405).

The original scope looks complete on current `main`. Please close this issue if there is no follow-up scope to add.

PR #394, opened 2026-09-19
feat(retrieval): find documents similar to a stored version

rodboev:
Docbank can find files similar to a selected document version from the web app, TUI, or CLI. It uses stored embedding rows, so it needs no query text or provider call.

The daemon enforces live, current, and source-fence checks, excludes the selected document, groups identical content with a count, and returns typed unavailable coverage when the embedding is missing. A separate route keeps ordinary text search unchanged.

Today, the workaround is a descriptive semantic query that sends text to the embedding provider. This slice adds the route, generated clients, CLI, web, TUI, and docs without new configuration. It follows [the maintainer direction](https://github.com/kenn-io/docbank/issues/392#issuecomment-5744241558). The screenshots use a temporary synthetic vault with no private corpus data. Refs #392

| Web result action | TUI similar view |
| --- | --- |
| The actual Docbank web application | The actual Docbank TUI |


wesm:
## Review summary

Reviewed the full diff against `main` (39 files). Build, `go vet`, `golangci-lint`, the Go suites for every touched package, `svelte-check`, the frontend unit tests, and both generated clients (`openapi.yaml`, `client.gen.go`, `docbank.ts`) were verified in sync before and after the fixes.

**Findings: 0 P1, 0 P2, 4 P3, 3 P4.** All four P3 findings are fixed in 6e258057; the P4 items are noted only.

### Fixed (P3)

- **Duplicate representative ignored the shared order on ties** (`internal/store/search_similar.go`). Identical content normally shares vector rows, so group members tie on score almost always. The tie-break then picked the lowest node ID, overriding the `DuplicateRepresentativeOrder` the query already applied. Similar results and the duplicates view could name different files for the same content group. Ties now keep the first row in representative order.
- **Wrong error class for an invalid limit** (`internal/processing/similar.go`). An out-of-range limit returned `ErrInvalidProcessingSourceFence`. Unreachable over HTTP because the schema bounds it, but embedded callers got a misleading error. Now a plain limit error, matching `Search`.
- **Cancellation check per scored pair** (`internal/vectorindex/generation.go`). `ctx.Err()` ran inside the innermost loop, up to a million mutex acquisitions per request. Moved to once per source row; the retrieval cancellation test still passes with its check count updated.
- **Screenshots README opened with a PR-only capture** (`frontend/screenshots/README.md`). The paragraph sat above the harness overview. Moved it to the bottom alongside the other PR-only cases, with a run command in the same shape.

### Noted, not changed (P4)

- `AcquireSimilarSearchAuthority` runs the source eligibility query twice before the fenced transaction runs it a third time. The first pass is load-bearing (it turns a never-processed profile into `unavailable` instead of a profile-lookup error, and the test pins that order). The pre-lease pass only changes which error wins when the head exists but the index is absent. Left as is.
- The web drawer picks `embedding_bindings[0]` client-side while the TUI lets the server choose; the server also picks the first binding, so they agree today.
- The `processingScope` expression is duplicated in `App.svelte`; two copies is under the threshold for extracting a helper.

Quality pipeline passes after the fixes: build, vet, lint, Go tests for the touched packages, `svelte-check`, and the frontend unit tests (73 files, 681 tests).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #395, opened 2026-09-19
feat(processing): add opt-in reranking to document search

rodboev:
Docbank search can now rerank results through a configured ZeroEntropy or Cohere provider when you pass `--rerank` or set `rerank` on `POST /api/v1/search`. Reranking stays within the caller's exact source-version set and is off by default.

Operators review the reranking provider, query text, and excerpt disclosure in the profile's plan. Approving that plan grants its configured operations together. Adding reranking changes the plan but leaves existing grants valid for the operations they already cover.

Search uses bounded excerpts from retained text and reports whether reranking was applied, skipped, or degraded. The configured failure policy either keeps the original ordering or fails the request. Missing local text is reported before calling the reranking provider.

This wires the existing reranking stage from [PR #221](https://github.com/kenn-io/docbank/pull/221) into daemon search. Web and TUI reranking controls and similar-file reranking are outside this change.

Refs [#392](https://github.com/kenn-io/docbank/issues/392#issuecomment-5744241558), slice 1 (wire reranking into daemon search).


wesm:
## Review of `a924cebf`

Findings: 0 P1, 1 P2, 3 P3, 4 P4. None fixed here: the head lives on a fork I cannot push to. The `content_fts` finding from roborev is resolved by `a924cebf`; I confirmed with a throwaway store test that after a real `plain-text` extraction the vault has zero `content_fts` rows and the original-file excerpt still resolves from `extracted_text`.

### P2

- **Configuring reranking forces re-consent and bundles the grant** (`internal/processing/service.go:1025`, `:1036`). Adding a `[processing_profiles.<name>.reranking]` section appends a `query_text_and_excerpt` request to every profile consent set. Existing plans then report consent `required` until re-granted, even for operators who never pass `--rerank`, and the only grant path grants embedding and reranking together. The docs describe this as a "separate" grant, but there is no separate decision. If this is intended, `docs/configuration.md` should state that adding the section invalidates active consent until the plan is re-reviewed.

### P3

- **User-facing search guide is silent on reranking.** `docs/usage/search.md` owns the consent walkthrough that `cli-reference.md` links to for `--rerank`, but it never mentions the option, its extra grant, or the receipt.
- **Empty excerpts degrade as `malformed_output`** (`internal/retrieval/providers.go:280`). That cause is documented as a provider output problem; here it means the local candidate had no text. The cause list is closed, so either document the extra meaning in `http-api.md` or leave it. Adding a new cause is not worth the client churn.
- **`documentSearchFlagsChanged` reads the global** (`cmd/docbank/search.go:217`) while its siblings use `cmd.Flags().Changed(...)`. Same behavior today, inconsistent shape.

### P4

- `internal/retrieval/media_evidence.go:97`: the `int64` overflow guard is unreachable for a sum of in-memory string lengths. A plain sum reads better.
- `internal/retrieval/search_stages_test.go:340`: stray `t.Logf("configured=7 ...")` diagnostic.
- `cmd/docbank/reranking_runtime.go:77`: the `default` error omits the profile name and is unreachable after config validation.
- `internal/config/config.go:767`: "between 1ns and 5m" reads oddly; "positive and at most 5m" matches the docs.

### Verification

`go test -tags fts5` passes for `internal/retrieval`, `internal/config`, `internal/daemonconn`, `internal/processing`, `internal/api`, `internal/store` (targeted), and `cmd/docbank`. `golangci-lint` reports 0 issues on the touched packages.

PR #396, opened 2026-09-19
feat(document): add TypeSafe Jev reranking provider

rodboev:
The reusable TypeSafe Jev provider client is ready for later integration. It asks one relevance question per candidate, and a retrieval adapter feeds its scores into the existing reranking contract.

The client sends TypeSafe the query and bounded excerpts. It strips document identities and evidence references, pins egress to api.typesafe.ai:443, bounds request work, and records provider token usage without retaining private text. Per-candidate and batched request shapes support the later evaluation.

A gated capture and replay test records and verifies both real TypeSafe wire shapes. Daemon search does not rerank through TypeSafe yet. Search wiring and the ZeroEntropy and Cohere comparison remain separate work.

Refs [#392](https://github.com/kenn-io/docbank/issues/392#issuecomment-5744241558), slice 2 (Jev provider)


rodboev:
## Proof report

Fixed the response handling in `document/typesafe/client.go` and added `TestRerankClassifiesStatusBeforeReadingBody`. The regression response is HTTP 429 with `Retry-After: 7`; its body fails if read. The test verifies the HTTP 429 status error, transient classification, 7 second retry delay, body unread, and body close.

Exact checks and output:

```text
go test -tags fts5 ./document/typesafe
ok  	go.kenn.io/docbank/document/typesafe	0.033s

go test -tags fts5 -race ./document/typesafe
ok  	go.kenn.io/docbank/document/typesafe	1.086s

go vet -tags fts5 ./document/typesafe
[no output]

gofmt -d document/typesafe/client.go document/typesafe/client_test.go
[no output]
```

wesm:
Pushed a small follow-up so the batched defaults can accept a full set of candidates. The old limit was 256 KiB, but 100 candidates at 4 KiB each already need about 400 KiB before JSON adds anything. Batched requests now default to 4 MiB, with room for escaped characters. Explicit request limits still apply.

The client now rejects non-JSON responses before reading the body. This avoids unnecessary reads and keeps a body-read failure from hiding the invalid response type.

I also replaced a duplicate nil-check helper with the existing shared helper. Regression tests cover full-size batches, explicit smaller limits, and rejecting non-JSON responses without reading their bodies.

The full Go test suite passed with both SQLite implementations. Focused race tests, vet with the capture tag, lint, and the commit checks also passed.

PR #399, opened 2026-09-21
Add CSV search exports and saved searches

salmonumbrella:
## What changed

- Search exports an eight-column CSV with each query's date range and direct, family, and unique counts. Each export freezes its source scope, date decisions, and coverage.
- The web app keeps the latest 100 saved searches per vault. Reuse one with new collections or dates to run against current data; prior counts and date choices stay with the original export.
- An evidence ZIP captures calculation inputs and hashes for later verification. The CLI and API expose the same export.

## Why

Search counts change when sources or processing change. Saving the request makes a later rerun quick, while a frozen export makes the CSV reproducible.

## Usage

Open **Search exports** in the web app, choose sources and cutoff dates, then select **Create export**. In **Saved searches**, choose **Use as draft** to rerun a request. For the CLI, run `docbank search-export create --input request.json --output export.zip`; use `docbank search-export csv export.zip --output counts.csv` to extract the CSV.

| Search export | Saved searches |
| --- | --- |
| Rendered search export with counts | Rendered saved searches with two source scopes |


wesm:
I checked the review findings against the code and pushed fixes in 9643d3c0 and baa865f3.

- Completed exports now release their temporary text allocations. Repeated exports no longer keep charging discarded text against the shared memory budget.
- Searches use the chosen processing profile. Offline ZIP verification recomputes coverage as well as counts and rejects inconsistent totals or unknown coverage states.
- Date review now requires the missing date and timezone before submission, and only offers actions supported by the evidence. Invalid input is separate from stale evidence. Oversized native or processed text returns a resource-limit error; exceeding the date-candidate limit also identifies the affected document.
- Unavailable export routes return an error instead of panicking, shutdown cleans up canceled builders, and opening a vault checks the new history table. Incomplete email attachment inventories remain visible, including when an attachment has been replaced.
- Collection exports filter documents in SQL, and ZIP creation indexes evidence instead of repeatedly scanning it. I removed the unused embedded export implementation and the unrelated frontend worker limit.

For the product overlap, I chose **Recent exports**: this is automatic history of the latest 100 runs, while named searches stay in **Saved queries and highlights**. The web wording, CLI, `/api/v1/search-exports` routes, downloaded filenames, and ZIP format name now use Search exports. The PR description includes fresh screenshots from the actual app using synthetic documents.

The new guide documents requests, date choices, count meanings, retained history, limits, and what offline verification proves. Public report types also have documentation comments.

I kept the second count calculation during ZIP creation because it checks results supplied by callers before publishing them. Reports still have bounded size and preparation time; this does not promise that every 50,000-document vault fits those limits. I left the original commit history intact.

Both complete Go suites passed (CGO and pure Go), followed by the final store regressions in both modes. Focused race checks, all 697 frontend tests and the frontend build, the strict docs build, client regeneration, the real-app screenshot test, and `prek` passed. The reported Windows-only glmocr failure was not reproduced or classified as a flake on this Linux host.

